PR #35, opened 2015-01-04
Creation of Policy with wizard

shlomizadok:
1) Create a wizard for new policy
2) Attach policy attributes (period, weekday, name) to selected hostgroup.


app/controllers/scaptimony_policies_controller.rb
```diff
@@ -8,8 +8,11 @@ def model_of_controller
   end
 
   def index
-    @policies = resource_base.search_for(params[:search]).includes(:scap_content, :scap_content_profile)
-    if Scaptimony::ScapContent.all.count == 0 && @policies.count == 0
+    @policies = resource_base.
+                  search_for(params[:search], :order => params[:order])
```
ohadlevy: nitpick: imho its more readable if the dot is at the end of the previous line.

app/controllers/scaptimony_policies_controller.rb
```diff
@@ -27,16 +30,24 @@ def parse
 
   def create
     @policy = ::Scaptimony::Policy.new(params[:policy])
-    if @policy.save
-      process_success
+    if @policy.current_step.blank? && @policy.save
+        process_success :success_redirect => scaptimony_policies_path
```
ohadlevy: nitpick: whitespace

app/controllers/scaptimony_policies_controller.rb
```diff
@@ -27,16 +30,24 @@ def parse
 
   def create
     @policy = ::Scaptimony::Policy.new(params[:policy])
-    if @policy.save
-      process_success
+    if @policy.current_step.blank? && @policy.save
```
ohadlevy: imho, its more readable to have a specific method instead of blank, e.g. something like
```ruby
@policy.completed_wizard? && policy.safe
```

app/helpers/scaptimony_policies_helper.rb
```diff
@@ -4,4 +4,22 @@ def profiles_selection
     return @policy.scap_content.scap_content_profiles unless @policy.scap_content.blank?
     return []
   end
+
+  def submit_or_cancel_policy(f, overwrite = nil, args = { })
+    args[:cancel_path] ||= send("#{controller_name}_path")
+    content_tag(:div, :class => "clearfix") do
+      content_tag(:div, :class => "form-actions") do
+        text    = overwrite ? overwrite : _("Submit")
```
ohadlevy: please simplify, override etc is not relevant here.

app/models/concerns/foreman_openscap/policy_extensions.rb
```diff
@@ -39,8 +93,57 @@ def used_organization_ids
...
+    private
+
+    def ensure_needed_puppetclasses
+      unless Puppetclass.find_by_name(SCAP_PUPPET_CLASS)
+        errors[:base] << _("Required Puppet class %{class} is not found, please ensure it imported first.") % {:class => SCAP_PUPPET_CLASS}
+      end
+    end
+
+    def assign_policy_to_hostgroups
+      if hostgroups
```
ohadlevy: wont hostgroups always have a value? maybe hostgroups.any?

app/views/scaptimony_policies/steps/_create_policy_form.html.erb
```diff
@@ -0,0 +1,7 @@
+<% display = @policy.current_step == step ? 'block' : 'none' %>
```
ohadlevy: you can use the css class hide instead.

app/views/scaptimony_policies/steps/_hostgroups_form.html.erb
```diff
@@ -0,0 +1,8 @@
+<% display = @policy.current_step == step ? 'block' : 'none' %>
```
ohadlevy: also, sounds like an helper :)

app/views/scaptimony_policies/steps/_locations_form.html.erb
```diff
@@ -0,0 +1,15 @@
...
+<% html_options ||= {} %>
+<% html_options[:location] ||= {} %>
+<% obj = @policy %>
+<% display = @policy.current_step == step ? 'block' : 'none' %>
+<% if show_location_tab? %>
+  <div class="tab-pane" id="locations" style="display: <%= display %>">
+    <%= wizard_header @policy.step_index, *@policy.steps %>
+    <%= location_selects f, obj.used_or_selected_location_ids,
+                         {:disabled => obj.used_location_ids }.merge!(select_options),
+                         {'data-descendants' => obj.children_of_selected_location_ids.to_json,
```
ohadlevy: nitpick - if you change to 
```ruby
{data => {...}}
```
then rails does to_json for you automatically.

app/views/scaptimony_policies/steps/_scap_content_form.html.erb
```diff
@@ -0,0 +1,21 @@
...
+               {:label => _("SCAP Content"),
+                :onchange => 'scap_content_selected(this);',
+                :'data-url' => method_path('scap_content_selected')}
+  %>
+  <span id="scap_content_profile_select">
+    <%= select_f f, :scap_content_profile_id, profiles_selection, :id, :title,
+                 {:selected => @policy.scap_content_profile_id,
+                  :include_blank => _("Default XCCDF profile")},
+                 {:label => _("XCCDF Profile"),
+                  :disabled => profiles_selection.empty? ? true : false,
```
ohadlevy: you can drop the true: false

app/views/scaptimony_policies/_form.html.erb
```diff
@@ -37,10 +38,16 @@
...
     </div>
     <div class="tab-pane" id="scap_schedule">
-      <%= text_f(f, :period) %>
-      <%= text_f(f, :weekday) %>
+      <%= select_f(f, :period, %w[Weekly Monthly], :downcase, :to_s, :include_blank => _('Choose period')) %>
+      <%= select_f(f, :weekday, Date::DAYNAMES, :downcase, :to_s, :include_blank => _('Choose weekday')) %>
     </div>
     <%= render 'taxonomies/loc_org_tabs', :f => f, :obj => @policy %>
+    <div class="tab-pane" id="hostgroups">
+      <% select_options ||= {} %>
```
ohadlevy: it does look like select_options is used anywhere? leftover?

app/helpers/scaptimony_policies_helper.rb
shlomizadok: It is relevant: It changes the text to "Next". See https://github.com/OpenSCAP/foreman_openscap/pull/35/files#diff-4951ee738829dd0659c417337012b434R9

app/helpers/scaptimony_policies_helper.rb
```diff
@@ -4,4 +4,26 @@ def profiles_selection
     return @policy.scap_content.scap_content_profiles unless @policy.scap_content.blank?
     return []
   end
+
+  def submit_or_cancel_policy(form, overwrite = nil, args = { })
+    args[:cancel_path] ||= send("#{controller_name}_path")
+    content_tag(:div, :class => "clearfix") do
+      content_tag(:div, :class => "form-actions") do
+        text    = overwrite ? overwrite : _("Submit")
```
ohadlevy: I mean that the overwrite is not needed, this is a copy paste from the app layout in foreman, which deals with overwrites for host creation (which should not have been global to start with).

ohadlevy: I see, please ignore.

app/assets/stylesheets/foreman_openscap/policy.css.scss
```diff
@@ -0,0 +1,7 @@
+.show-pane {
+  display: block;
+}
+
+.hide-pane {
+  display: none;
```
ohadlevy: why not use the hide css class?

app/assets/stylesheets/foreman_openscap/policy.css.scss
```diff
@@ -0,0 +1,7 @@
+.show-pane {
+  display: block;
```
ohadlevy: and show css class ?

isimluk:
The following search/scope queries needs to be amended after scaptimony_policy_hostgroups is introduced:
```ruby
Host.search_by_policy_name
Host.search_by_missing_arf
Scaptimony::Asset.policy_reports_missing
```

Previously we had only one possible mapping between an AuditableAsset and a Policy. Now two possible mappings exists. The rails are not particularly good in this, so I would create UNION view for asset--policy mapping.

Note that the we have this Asset != Host indirection, because we will need to audit containers in future.

ohadlevy:
@isimluk what do you mean by audit containers?

isimluk:
Container (i.e. docker) environment can be audited with OpenSCAP same way as the host machine. In fact there are already deployments using OpenSCAP to audit containers. SCAPtimony should be able to store and process the arf_reports from containers same way as from hosts. Hence the Asset class exists. </ot>

isimluk:
Looking here. I will retest this on top of master and merge. We will amend aforementioned queries later on.

isimluk:
Ok, so this will need to have a few amendments with regards to polymorphic asset. I'll make it.

isimluk:
Merged as 71c5e1ada60063227e034323af8392d981ee1b2b and amended in 8acba54deeed8fedf0a662f20ee108cc86d62823.

isimluk:
There is still some unresolved issue with the hostgroup assignment. I hope to fix it later this night.

I have decided to merge this before we have all the issues addressed. As there is value in having this in master finally. @ares plans to make some changes related to policy assignment.

Thanks @shlomizadok and @ohadlevy for the Policy creation guide!!

Issue #38, opened 2015-01-07
taxonomy of ARF report ?


isimluk:
See  c911d65fc965a45c142e8a735ba1c698b030e255 and ed8e64d97081edd25c0d1e3909acdcbe97f08a3a.

Me and @ares came to the conclusion that ArfReport should always be in the same org/location as the Host is. Hence, we perhaps don't need to have separate taxonomy. We just need to create a default_scope for ArfReport that would reflect current org/loc.

Thoughts?

shlomizadok:
I think it is okay. Anyhow, if you assign policy to a host it would be in the context of the taxonomies.
Added a default_scope to ArfReport in #39

isimluk:
The problem is when a host is migrated to different taxonomy, then the /compliance/reports break, because the previous organization will not see the report's host (the host would be nil I think).

I'll leave this question open for now. But perhaps it is not a top priority.

shlomizadok:
This was actually fixed in #39. Can you re-test?
For safety, maybe we can refactor to ```<%= name_column(arf_report.host) if arf_report.host %>```

isimluk:
Ok, will retest. Thanks!

isimluk:
Right. I was not able to break /compliance/reports page. I was not able to change the org/loc of existing host.

ares:
What about scap content taxonomies? If we remove taxonomies from policy we should do the same for scap contents but here I'm not sure whether it makes sense. In multitenancy setup I see use case where you want scoped scap content. And then there's a question whether policy should "inherit" taxonomies from content or we should even allow subset of content taxonomies.

If we don't have easy answers here I'm thinking of leaving taxonomies as it is now. Warning: I'm still not confident with all details of openscap plugin so I may be completely wrong.

shlomizadok:
I am in favor of making scap_content system-wide and not bind scap_content
to taxonomies. I think it should be system-wide, because you can set only
*one* scap_content from a file (e.g., you can't create 'fedora'
scap_content twice - one for org x and one for org y). Therefore, I think
that scap_content should be without taxonomies at all.

2015-01-08 13:44 GMT+02:00 Marek Hulán <notifications@github.com>:

isimluk:
@shlomizadok , I agree with @ares that in multi-tenant environment: there might contents that should not be shared.

@ares it is interesting idea to inherit taxonomy from content to the policy and keep it the same.

[...]

PR #39, opened 2015-01-07
added default scope with taxonomies

shlomizadok:
@isimluk -- the reason we haven't seen reports by taxonomies is because I've forgot to add the default scope with taxonomies...


isimluk:
Haven't tried yet, but should we order in "desc"?

Could you please take a look at #38. Well, that is rather a question, but it is related.

shlomizadok:
Added desc.
As for #38 - will answer there...

isimluk:
Thank You!

PR #46, opened 2015-01-15
fixes to align to foreman 1.7-stable

shlomizadok:
1. returned to ```'Host::Base'```, instead of ```'::Host::Base'``` - reason for that: with ```'::Host::Base'```, you will get no results for host finders (e.g: ```Host.last.arf_reports```, or ```Host.last.asset```, etc.) which also yields "no compliance reports" for hosts with reports on hosts#index deface

2. ```:no_whitespace => true``` is not merged in foreman-1.7-stable so we can't use it. refactored to ```:format => { without: /\s/ }```, which does the same.

### DIFF
--- a/app/models/concerns/foreman_openscap/arf_report_extensions.rb
+++ b/app/models/concerns/foreman_openscap/arf_report_extensions.rb
@@ -15,7 +15,7 @@ module ForemanOpenscap
     extend ActiveSupport::Concern
     include Taxonomix
     included do
-      has_one :host, :through => :asset, :as => :assetable, :source => :assetable, :source_type => '::Host::Base'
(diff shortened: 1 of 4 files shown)


isimluk:
I have made the change to `::Host::Base` to be get host from ArfReport like: `Scaptimony::ArfReport.find(1).host`. See 00ae7b704e9e77884cc43087bb1493ac19a2b6b8. Does that work after your change?

shlomizadok:
@isimluk yeah, re-tested: Scaptimony::ArfReport.first.host (or Scaptimony::ArfReport.find(1).host) returns the host.

isimluk:
Sight. This is embarrassing. It works, but I don't understand. It haven't worked before. 

Thanks for correction.

isimluk:
(and merged)

PR #64, opened 2015-01-20
Change API path to /compliance

isimluk:
Fixes: #44.

PR #65, opened 2015-01-20
[Do not merge] Client sends policy_id instead of policy_name

### LINKED ISSUE #45: api should take policy_id instead of policy_name
The new foreman_scap_client sends the ID. So the foreman_proxy-openscap forwards policy_id, but scaptimony expects policy_name.

We need to merge #42 first -- otherwise there will be merge conflict.

shlomizadok:
Changes from policy_name to policy_id.
Merge with OpenSCAP/scaptimony#16
Merge when client change will be merged.
Fixes #45

### DIFF
--- a/app/controllers/api/v2/openscap/arf_reports_controller.rb
+++ b/app/controllers/api/v2/openscap/arf_reports_controller.rb
@@ -21,9 +21,9 @@ module Api
 
         add_puppetmaster_filters :create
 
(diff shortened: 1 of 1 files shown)


isimluk:
We have smart_proxy_openscap and scaptimony parts merged. We just need this. But it conflicts with my openscap->compliance move. I have added your commit on top of my PR: #64. Please review.

PR #66, opened 2015-01-21
fixes #62 - Adds a notice that the scap_content needs to be uploaded to ...

### LINKED ISSUE #62: Notify the user that the scap content needs to be deployed to the host
While creating a new policy (on the scap content step, notify the user that the scap content needs to be deployed to the host) -- [with link to explanation how to?]

shlomizadok:
...the hosts


isimluk:
For production, I would not link to the upstream project page. But for upstream foreman_openscap I think this is acceptable.

Thanks!

PR #72, opened 2015-01-29
Require specific SCAPtimony version.

isimluk:


foreman_openscap.gemspec
```diff
@@ -15,5 +15,5 @@ Gem::Specification.new do |s|
   s.test_files = Dir["test/**/*"]
 
   s.add_dependency 'deface', '~> 0'
-  s.add_dependency 'scaptimony', '~> 0'
+  s.add_dependency 'scaptimony', ['>= 0.3.0', '< 0.4.0']
```
shlomizadok: Won't ```~> 0.3``` do the job?

isimluk: In fact, the `~> 0.3.0` is needed. The `~> 0.3` has different semantics. That only shows it, how this notation is confusing.

ares: definitely for non-rubyist, IMHO rubyists will find `['>= 0.3.0', '< 0.4.0']` weird and will start to think why you used this... anyway it's a small detail

shlomizadok:
excellent. merging in 3, 2, 1....

PR #75, opened 2015-02-05
Bind puppetmaster fqdn instead of full URI.

isimluk:
Otherwise the foreman_scap_client will attempt to connect to
https://http//foreman17.local.lan

### DIFF
--- a/app/models/concerns/foreman_openscap/policy_extensions.rb
+++ b/app/models/concerns/foreman_openscap/policy_extensions.rb
@@ -224,7 +224,8 @@ module ForemanOpenscap
       puppetclass.class_params.where(:override => true, :key => SERVER_CLASS_PARAMETER).each do |override|
         if hostgroup.puppet_proxy && (url = hostgroup.puppet_proxy.url).present?
           lookup_value = LookupValue.where(:match => "hostgroup=#{hostgroup.to_label}", :lookup_key_id => override.id).first_or_initialize
-          lookup_value.update_attribute(:value, url)
+          puppet_proxy_fqdn = URI.parse(url).host
+          lookup_value.update_attribute(:value, puppet_proxy_fqdn)
         end
       end
     end


shlomizadok:
I think we should actually change the foreman client behavior, because the "https://" is always hard coded there (https://github.com/OpenSCAP/foreman_scap_client/blob/master/lib/foreman_scap_client/client.rb#L112) -- we should support both "http://" and "https://" and the value from foreman proxy url is exactly that value. 
Speaking of which we should add a config to the proxy port  (may be 8443, but also 8000)

isimluk:
Agree. We can introduce `url` parameter for foreman_scap_client. Btw we already support custom port.

However, now there is a bug in foreman_openscap, it binds wrong value to existing parameter (param name is `server` not `url`.

shlomizadok:
I'm not sure I understand. This fix takes only the proxy host instead of full url, and adds it to the puppet class overrides, right? But why? Why can't we use the server parameter with the full url? the fix actually should be on the foreman_scap_client - and remove the "https" from there.
so on the server parameter we will have value like "https://proxy.example.com:8443", which is the exact url we need to add the compliance path and send data to.

isimluk:
Ok, I will reword in simple senteces.

(1) foreman_scap_client does not take $url parameter.
(2) foreman_scap_client takes parameter called $server
(3) foreman_scap_client takes parameter called $port
(4) $server is expected to be server identifier (fqdn or ip) not uri or url
(5) if you want to use uri, you will need to create $uri parameter in foreman_scap_client
(6) I don't think there is enough time to do another rebase of clients.

ohadlevy:
I tend to agree with @shlomizadok , @ares whats your take on this?

ares:
While I agree that we should be able to specify the whole url I agree with @isimluk that the change would affect foreman_scap_client, puppet module and foreman_openscap. So I'd postpone improving this to next version (either by migrating to $url or introducing other parameter e.g. $scheme). It shouldn't be hard to update, since we'll just regenerate foreman_scap_client config file using new puppet manifest). Btw other issue with current implementation (regardless this PR) is with sub-uri deployments of smart proxies (probably not common).

ohadlevy:
ok, so :+1: to merge, @isimluk ?
@shlomizadok I suggest another set of PR's so it gets in for the next stable/bugfix version?

shlomizadok:
merging under protest ;)

PR #78, opened 2015-02-12
Setting location & organization from host should be by id

shlomizadok:
An error will occur if trying to set ```location_ids = [host.location]``` should be by taxonomy_id


ares:
LGTM

PR #79, opened 2015-02-12
fixes #59 - disassociate policy from hosts

### LINKED ISSUE #59: as a user i would like to de-assoicate  a policy from a host

shlomizadok:
Fixes #59

### DIFF
--- a/app/controllers/scaptimony_policies_controller.rb
+++ b/app/controllers/scaptimony_policies_controller.rb
@@ -1,7 +1,7 @@
 class ScaptimonyPoliciesController < ApplicationController
   include Foreman::Controller::AutoCompleteSearch
   before_filter :find_by_id, :only => [:show, :edit, :update, :parse, :destroy]
-  before_filter :find_multiple, :only => [:select_multiple_hosts, :update_multiple_hosts]
+  before_filter :find_multiple, :only => [:select_multiple_hosts, :update_multiple_hosts, :disassociate_multiple_hosts, :remove_policy_from_multiple_hosts]
 
   def model_of_controller
     ::Scaptimony::Policy
@@ -73,13 +73,27 @@ class ScaptimonyPoliciesController < ApplicationController
   def update_multiple_hosts
     if (id = params['policy']['id'])
       policy = ::Scaptimony::Policy.find(id)
-      policy.assign_hosts @hosts
+      policy.assign_hosts(@hosts)
       notice _("Updated hosts: Assigned with compliance policy: #{policy.name}")
       # We prefer to go back as this does not lose the current search
       redirect_to hosts_path
     else
       error _('No compliance policy selected.')
       redirect_to(select_multiple_hosts_scaptimony_policies_path)
+    end
+  end
+
+  def disassociate_multiple_hosts; end
+
+  def remove_policy_from_multiple_hosts
+    if (id = params.fetch(:policy, {})[:id])
+      policy = ::Scaptimony::Policy.find(id)
+      policy.unassign_hosts(@hosts)
+      notice _("Updated hosts: Unassigned from compliance policy '%s'") % policy.name
+      redirect_to hosts_path
+    else
+      error _('No valid policy id provided')
+      redirect_to hosts_path
     end
   end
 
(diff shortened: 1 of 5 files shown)


app/controllers/scaptimony_policies_controller.rb
```diff
@@ -83,6 +83,17 @@ def update_multiple_hosts
     end
   end
 
+  def disassociate_multiple_hosts; end
+
+  def remove_policy_from_multiple_hosts
+    if (id = params['policy']['id'])
+      policy = ::Scaptimony::Policy.find(id)
+      policy.unassign_hosts(@hosts)
+      notice _("Updated hosts: Unassigned from compliance policy: #{policy.name}")
```
ares: I think this would be better
```ruby
_("Updated hosts: unassigned from compliance policy '%s'") % policy.name
```

app/controllers/scaptimony_policies_controller.rb
```diff
@@ -83,6 +83,17 @@ def update_multiple_hosts
     end
   end
 
+  def disassociate_multiple_hosts; end
+
+  def remove_policy_from_multiple_hosts
+    if (id = params['policy']['id'])
```
ares: two notes here
1. `params.fetch(:policy, {})[:id]` would be a slightly safer
2. it seems we don't handle else case anyway

app/models/concerns/foreman_openscap/policy_extensions.rb
```diff
@@ -126,6 +126,11 @@ def assign_hosts(hosts)
       assign_assets hosts.map &:get_asset
     end
 
+    def unassign_hosts(hosts)
+      hosts = hosts.map &:get_asset
+      self.asset_ids = (self.asset_ids - hosts.collect(&:id)).uniq
```
ares: minor optimization
```ruby
host_asset_ids = hosts.map { |h| h.get_asset.id }
self.asset_ids = self.asset_ids - host_asset_ids
```

we can obtain ids in one cycle plus I think that uniq is not necessary since we just substract
also note that this will generate SQL query for each host, it may be better to either create get_assets(hosts) method somewhere or use Scaptimony::Asset directly here

app/views/scaptimony_policies/disassociate_multiple_hosts.html.erb
```diff
@@ -0,0 +1,7 @@
+<%= render 'hosts/selected_hosts', :hosts => @hosts %>
+<%= form_for :policy,
+             :url => remove_policy_from_multiple_hosts_scaptimony_policies_path(:host_ids => params[:host_ids]) do |f| %>
+    <%= selectable_f f, :id, [[_("Select Compliance Policy"), "disabled"],
```
ares: Just out of curiosity, I suppose we can't use :placeholder here because of the disabled values? Also if possible try to use `''` instead of `""` when there's no interpolation in strings.

app/controllers/scaptimony_policies_controller.rb
shlomizadok: Changed. Thanks.

ares:
Wow, that was quick :-) my last concern is the SQL query for each host in `def unassign_hosts(hosts)`.

shlomizadok:
@ares - thanks for the review, I have applied the changes.
Regarding your concern about the SQL query: the hosts objects are already passing, so the SQL is only querying the Scaptimony::Asset (based on the optimization you have proposed).
So it looks good to me.

ares:
This line 
```ruby
host_asset_ids = hosts.map { |h| h.get_asset.id }
``` 
runs get_asset for each host, which means that
```ruby
Scaptimony::Asset.where(:assetable_type => 'Host::Base', :assetable_id => id).first_or_create!
``` 
(content of get_asset) generates new SQL query for every host. I think we should simply write it like

```ruby
host_asset_ids = Scaptimony::Asset.where(:assetable_type => 'Host::Base', :assetable_id => @hosts.map(&:id)).pluck(:id)
```
which would load all ids in one SQL query.

shlomizadok:
Oh, I see your point. done :+1:  thanks.

ares:
Seems good to me (untested though).

PR #80, opened 2015-02-16
fixes #22 - Adds vreification for openscap on smart_proxy

### LINKED ISSUE #22: The scap proxy is not ferified when not puppet (or chef) installed on the proxy
When authorizing the smart proxy, the code used (https://github.com/OpenSCAP/foreman_openscap/blob/master/app/controllers/api/v2/openscap/arf_reports_controller.rb#L22) require the proxy to have 'Puppet' or a 'Chef Proxy' as it's features.

This means that when the scap proxy was installed as standalone, without a puppet master (or chef proxy), the calls to the Foreman would be refused.

The `foreman_abrt` plugin has the same issue.

One possible solution would be to extract the list of features to be checked into a method, and override this method from within the `ArfReportsController`

shlomizadok:
@inecas - will you kindly take a look?

### DIFF
--- a/app/controllers/api/v2/compliance/arf_reports_controller.rb
+++ b/app/controllers/api/v2/compliance/arf_reports_controller.rb
@@ -19,7 +19,7 @@ module Api
         include Api::Version2
         include Foreman::Controller::SmartProxyAuth
 
-        add_puppetmaster_filters :create
+        add_smart_proxy_filters :create, :features => 'Openscap'
 
         api :POST, "/arf/:cname/:policy_id/:date", N_("Upload an ARF report")
         param :cname, :identifier, :required => true


iNecas:
Since the change needed on the Foreman side was added recently (not even in 1.7 https://github.com/theforeman/foreman/pull/1975), so the requirements on the foreman version should be `>= 1.8` https://github.com/OpenSCAP/foreman_openscap/blob/master/lib/foreman_openscap/engine.rb#L35
(or trying to get the required change backported to Foreman 1.7)

shlomizadok:
moved to 0.4.0, as it requires foreman-1.8 +

isimluk:
I will probably create maintenance branch for foreman 1.7. And then I will merge this to master.

PR #82, opened 2015-02-19
Create a host view

shlomizadok:
host_policies_reports

Creates a host view for all related reports and policies. Accessible via hosts#show


isimluk:
Merged as 328a7e02d50b3d6c5dac6f8432a35d90c9c9a220.

Thank You!

PR #83, opened 2015-02-25
Remove Rakefile from gemspec

isimluk:
Rakefile has been removed from repository
by 1f7755c721808b4e5f318c829845d50103941f1d

Addressing error while building gem:
WARNING:  See http://guides.rubygems.org/specification-reference/ for help
ERROR:  While executing gem ... (Gem::InvalidSpecificationException)
    ["Rakefile"] are not files


shlomizadok:
Great catch. Thanks.

PR #84, opened 2015-02-26
Adds validation to scap_content_profile_id

shlomizadok:
If scap_content_profile_id is blank, it will raise an error on the "view guide" page, and may run incorrectly on the client.


isimluk:
First, it should go to scaptimony, not to the concern.

Second, it is not correct. Having no profile is correct use-case we need to support.

The client tool already supports this. What is the error message when generating guide?

shlomizadok:
@isimluk -- Agreed. The issue is that the guide is not generated and we get an erb missing error.
Closing here and pushing a small fix to Scaptimony

PR #86, opened 2015-03-01
create class param for port

shlomizadok:
**Merge with https://github.com/OpenSCAP/puppet-foreman_scap_client/pull/5**

Reason:
Port may not always be 8443, as was hardcoded. It may be, 8000 (no ssl) or 9090 (capsule)


isimluk:
Merged as cc1db34f86a99b1df2faae6df8e092fa5169f336

Thanks!

PR #96, opened 2015-03-26
Enable katello

shlomizadok:
File ```lib/foreman_openscap/helper.rb``` is now testing if cname is a uuid or a host name.

File ```app/models/concerns/foreman_openscap/policy_extensions.rb``` Adds overrides to the Puppet module to use Katello certificates. This is now ideal, as we take to assumption that the host will also have a content host.

@ohadlevy, @ares, @stbenjam - please review as well. (Thanks!)

### DIFF
--- a/lib/foreman_openscap/helper.rb
+++ b/lib/foreman_openscap/helper.rb
@@ -10,8 +10,25 @@
 
 module ForemanOpenscap::Helper
   def self.get_asset(cname, policy_id)
-    asset = Host.find_by_name!(cname).get_asset
+    asset = find_host_by_name_or_uuid(cname).get_asset
     asset.policy_ids += [policy_id]
     asset
+  end
+
+  private
+
+  def self.find_host_by_name_or_uuid(cname)
+    if defined?(Katello::System)
+      host = Host.includes(:content_host).where(:katello_systems => {:uuid => cname}).first
+      host ||= Host.find_by_name(cname)
+    else
+      host = Host.find_by_name(cname)
+    end
+    unless host
+      Rails.logger.error "Could not find Host with name: #{cname}"
+      Rails.logger.error "Please check that Content host is linked to Foreman host" if defined?(Katello::System)
+      raise ActiveRecord::RecordNotFound
+    end
+    host
   end
 end


app/models/concerns/foreman_openscap/policy_extensions.rb
```diff
@@ -244,5 +251,28 @@ def populate_overrides(puppetclass, hostgroup)
...
+        end
+        certificate_param.override = true
+        case override_parameter
+          when CA_FILE_PARAMETER
+            certificate_param.default_value = '/etc/rhsm/ca/katello-server-ca.pem'
+          when HOST_CERTIFICATE_PARAMETER
+            certificate_param.default_value = '/etc/pki/consumer/cert.pem'
+          when HOST_PRIVATE_KEY_PARAMETER
+            certificate_param.default_value = '/etc/pki/consumer/key.pem'
+        end
```
stbenjam: Oh, can you not read the rhsm.conf on the client uploading tool instead of sending them via Puppet?

Depending on when the system was registered these could be different.... at one point, the file wasn't named /etc/rhsm/ca/katello-server-ca.pem.

app/models/concerns/foreman_openscap/policy_extensions.rb
```diff
@@ -244,5 +251,28 @@ def populate_overrides(puppetclass, hostgroup)
...
+    def set_certificates_overrides(puppetclass)
+      [CA_FILE_PARAMETER, HOST_CERTIFICATE_PARAMETER, HOST_PRIVATE_KEY_PARAMETER].each do |override_parameter|
+        unless certificate_param = puppetclass.class_params.where(:key => override_parameter).first
+          errors[:base] << _("Puppet class %{class} does not have %{parameter} class parameter.") % {:class => SCAP_PUPPET_CLASS, :parameter => override_parameter}
+          return false
+        end
+        certificate_param.override = true
+        case override_parameter
+          when CA_FILE_PARAMETER
+            certificate_param.default_value = '/etc/rhsm/ca/katello-server-ca.pem'
```
stbenjam: Oh, can you not read the rhsm.conf on the client uploading tool instead of sending them via Puppet?

Depending on when the system was registered these could be different.... at one point, the file wasn't named /etc/rhsm/ca/katello-server-ca.pem.

ohadlevy: reading another system config files does not sound ideal either, can the installer / some other thing maintain another config file / links to the ca etc ?

stbenjam: Why not? They're authenticating with the RHSM certificates; they should use the canonical source for that information, which is `/etc/rhsm/rhsm.conf`.  katello-agent does the same, although that is more closely linked to rhsm.

For whatever reason, the certificate was renamed from `/etc/rhsm/ca/candlepin-local.pem` to  `/etc/rhsm/ca/katello-server-ca.pem`, so older clients may have the older cert location.

If you don't read from rhsm.conf, you'd be forcing users to regenerate all of their certificate RPMs and redeploy them with the newer names, or change all of their older clients some other way? I don't think that's a reasonable ask.

The client knows where its certificates are, why is it being sent from the Foreman server anyway?

ohadlevy: we still want to support both usage cases (puppet/rhsm certs). setting it
via puppet allows the user to override it in a ui for all clients as a
configuration directive....the other alternative is to make the client
"smarter" but I'm not 100% sure if its required, surely better error
messages on the on the client can help too...

stbenjam: Maybe you could have a boolean for using the rhsm to upload.

ohadlevy: actually, since the certs to use are defined by the proxy we use, i wonder
if there is no better way to detect which 'kind' of proxy and provide
default certs accordingly ?

lib/foreman_openscap/helper.rb
```diff
@@ -10,8 +10,24 @@
...
-    asset = Host.find_by_name!(cname).get_asset
+    asset = find_host_by_name_or_uuid(cname).get_asset
     asset.policy_ids += [policy_id]
     asset
   end
+
+  private
+
+  def self.find_host_by_name_or_uuid(cname)
+    if defined?(Katello::System)
```
ohadlevy: missing tests :)

lib/foreman_openscap/helper.rb
```diff
@@ -10,8 +10,24 @@
...
+      host = Host.includes(:content_host).where(:katello_systems => {:uuid => cname}).first
+      host ||= Host.find_by_name(cname)
+    else
+      host = Host.find_by_name(cname)
+    end
+    unless host
+      Rails.logger.error "Could not find Host with name: #{cname}"
```
ohadlevy: I would add a comment here, that this can cause where there is a content host which is not linked into a foreman host.

lib/foreman_openscap/helper.rb
shlomizadok: As all of this project... :)
Some were created in https://github.com/shlomizadok/foreman_openscap/tree/host_form_scap_policy_proxy/test, but this branch is still work in progess

app/models/concerns/foreman_openscap/policy_extensions.rb
stbenjam: Why would you do it based on Proxy? If the config is being set from Foreman, you know if the Host object has a ContentHost and could tell it to use the RHSM certs in that case, but the paths to the certs should come from /etc/rhsm/rhsm.conf still.

ohadlevy: the proxy that runs scap might be configured with puppet certs vs katello
certs.

also, ATM the evaluation is not done per host (there was a desire not to
 define the ssl configuration per host in foreman lookup_values/smart class
parameters).

stbenjam: How would a proxy connected to Katello be using puppet certs??

You all can solve it however you want, but if you hardcode RHSM cert locations, you're going to run into problems.  That's my only comment about this PR.

ohadlevy: do you happen to know if subscription manager can output its configuration
(maybe use it as a fact instead?)

stbenjam: `subscription-manager config` but it's basically just catting /etc/rhsm/rhsm.conf

stbenjam: Actually, subscription-manager config may be preferable, because it doesn't the variable substitution you'd have to deal with in rhsm.conf,

   repo_ca_cert = /etc/rhsm/ca/katello-server-ca.pem

So, yea, that could be turned into a fact

shlomizadok:
@stbenjam @ohadlevy @ares, thank you all for the comments and help.
I have remove the hard coded puppet class params overrides.
Instead, I am testing which certificates to use on the puppet module.
Would love if you could take a look and review https://github.com/OpenSCAP/puppet-foreman_scap_client/pull/6 in addition to this PR.
Thanks!

lib/foreman_openscap/helper.rb
stbenjam: Or create an unmanaged host like uploading reports/facts does? https://github.com/theforeman/foreman/blob/develop/app/services/report_importer.rb#L69

shlomizadok: I am not sure as this is due to change when single host feature will be applied.
@ohadlevy - what says you?

stbenjam: Oh fair enough, probably not necessary.  

If a host is reporting openscap, it should have a foreman host because it came via Puppet. It's Katello's problem if this situation happens, that there's no corresponding foreman host.

shlomizadok:
@ares @isimluk, I believe this PR is also ready to merge.

ohadlevy:
@isimluk ping? any reason not to merge?

ares:
Thanks @shlomizadok, merging!

ares:
@shlomizadok actually before merge, could you squash changes in policy_extensions.rb so we don't have one commit adding stuff that is removed in second? I did review of the whole then realized that there are three commits and most of 1st and 2nd neutralize each other :-) I'd probably squash into one as this is one feature as a whole

shlomizadok:
@ares squashed. Thanks.

ares:
Merged, thanks!

PR #97, opened 2015-04-15
API to display policy xml content

shlomizadok:
Not sure if needs to merge to this branch.
Creates api to policy API.

**Should be merged also to master branch**
(may need some changes in routes.rb and  policies_controller.rb )

### DIFF
--- a/app/models/concerns/foreman_openscap/host_extensions.rb
+++ b/app/models/concerns/foreman_openscap/host_extensions.rb
@@ -26,7 +26,7 @@ module ForemanOpenscap
     end
 
     def combined_policies
-      combined = self.policies + self.hostgroup.policies
+      combined = self.hostgroup ? self.policies + self.hostgroup.policies : self.policies
       combined.uniq
     end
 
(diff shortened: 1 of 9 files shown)


app/models/concerns/foreman_openscap/host_extensions.rb
```diff
@@ -26,7 +26,7 @@ def policies_enc
     end
 
     def combined_policies
-      combined = self.policies + self.hostgroup.policies
+      combined = self.hostgroup ? self.policies + self.hostgroup.policies : self.policies
```
shlomizadok: This should be ported also to master branch

app/models/concerns/foreman_openscap/policy_extensions.rb
```diff
@@ -134,6 +134,7 @@ def to_enc
         'id' => self.id,
         'profile_id' => self.scap_content_profile.try(:profile_id) || '',
         'content_path' => "/var/lib/openscap/content/#{self.scap_content.digest}.xml",
+        'download_path' => "/compliance/policies/#{self.id}/content" # default to proxy path
```
shlomizadok: Added in https://github.com/OpenSCAP/puppet-foreman_scap_client/pull/11

shlomizadok:
@ares, @ohadlevy - Would love to have your take on this

isimluk:
This should first go to master branch. And only after that being backported to maintenance branch. This process called backward patching.

I agree that opposite approach (forward patching) has it merits. The problem is that I have been forced to start with backward paching for historical reasons.

shlomizadok:
#99 adds this functionality to master branch

isimluk:
This can be merged.

However, I would wait until it is clear that we need this in the maintenance branch.

Thanks!

PR #100, opened 2015-04-30
Fixes #10335 - support reports without host

ares:


ares:
First I was trying to make reports deleted upon host deletion, unfortunately dependent destroy does not work. This way reports remain in scaptimony and can be inspected even after host was deleted which might be useful.

isimluk:
Merged to master as 5bbba3eeb52c66d13e40ea99f247097efea539a7.
Merged to maint as 9da057360fb9e5daa4cd721956847af0b77ed5fe. 

Thanks!

PR #104, opened 2015-05-20
Fixes #10549 - show host name only if host exists

shlomizadok:
Needs to be backported to maint-foreman17 too


ares:
LGTM, not tested

ares:
oh yeah, it was a module so you it must be alias, I think you can squash both commits and it's ready

ares:
Works nicely, merging. Thanks @shlomizadok.

ares:
@shlomizadok would you mind CP to maint-foreman17?

shlomizadok:
Done

PR #105, opened 2015-05-20
add permissions to comply with 1.8+

shlomizadok:


app/controllers/api/v2/compliance/policies_controller.rb
```diff
@@ -1,6 +1,10 @@
 module Api::V2
   module Compliance
     class PoliciesController < ::Api::V2::BaseController
+      include Foreman::Controller::SmartProxyAuth
+
+      add_smart_proxy_filters :content
```
ares: I don't see this feature defined in smart_proxy_openscap, isn't content feature added by pulp node? Should we  have our own feature perhaps?

shlomizadok: content feature isn't defined by pulp. yet you are right - it should be similar to https://github.com/theforeman/foreman_openscap/blob/master/app/controllers/api/v2/compliance/arf_reports_controller.rb#L22
fixing. Thanks :+1:

ares:
LGTM

ares:
Tested, fixes the unauthorized issue when downloading content file, thanks @shlomizadok

PR #112, opened 2015-06-17
fixes #9950 - ensure scap content from file is created only once

shlomizadok:
@ares - this fixes when __same__ scap content is created (or usually fail) on each db:seed run


ares:
Fixes the issue for me, thanks @shlomizadok, merging!

PR #113, opened 2015-06-17
fixes #10677 - remove logging of scap content

shlomizadok:
with the help of @ehelms


ehelms:
:+1:

ares:
Much better, now we should silence the SQL log :-) Anyway that's a different story, merging this one! Thanks @shlomizadok

PR #119, opened 2015-08-03
Fixes #11132 - Showing spinner until a report is loaded

xprazak2:
We wait for a report to fully load and we show a spinner in the meantime.

### DIFF
--- a/app/views/scaptimony_arf_reports/show.html.erb
+++ b/app/views/scaptimony_arf_reports/show.html.erb
@@ -1,3 +1,12 @@
+<%= javascript 'foreman_openscap/load_report'%>
 <div class="row">
-  <iframe style="min-height: 800px" height="100%" width="100%" frameborder="0" src="<%= parse_scaptimony_arf_report_path(@arf_report) %>"></iframe>
+
+  <div id="loading">
+    <div class="col-md-4"></div>
+    <div class="col-md-4">
+      <%= image_tag('/assets/spinner.gif') %>
+      <%= _('Loading...') %>
+    </div>
+  </div>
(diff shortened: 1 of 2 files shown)


app/views/scaptimony_arf_reports/show.html.erb
```diff
@@ -1,3 +1,12 @@
+<%= javascript 'foreman_openscap/load_report'%>
 <div class="row">
-  <iframe style="min-height: 800px" height="100%" width="100%" frameborder="0" src="<%= parse_scaptimony_arf_report_path(@arf_report) %>"></iframe>
+
+  <div id="loading">
+    <div class="col-md-4"></div>
+    <div class="col-md-4">
+      <img  src='/assets/spinner.gif'> Loading...
```
shlomizadok: Translation

app/views/scaptimony_arf_reports/show.html.erb
```diff
@@ -1,3 +1,12 @@
...
+    </div>
+  </div>
+
```
shlomizadok: nitpick: extra line?

shlomizadok:
Tested. works great.
The "Loading" text should be extracted to translation.
Will merge when this is fixed. 
Thanks!

app/views/scaptimony_arf_reports/show.html.erb
shlomizadok: ```<img src...``` Should be moved to ```ruby <%= image_tag('/assets/spinner.gif'....``` - It will also be easier to add the translation this way

shlomizadok:
Thanks @xprazak2

PR #120, opened 2015-08-10
Fixes #11198 - apipie_documented_controllers defined

xprazak2:


lib/foreman_openscap/engine.rb
```diff
@@ -32,10 +32,16 @@ class Engine < ::Rails::Engine
                                        'foreman_openscap/policy.css']}}
     end
 
+    initializer 'foreman_openscap.apipie' do
+      Apipie.configuration.checksum_path += ['openscap/api']
```
shlomizadok: Should be
```ruby
Apipie.configuration.checksum_path += ['/compliance/']
```
(see [foreman_discovery](https://github.com/theforeman/foreman_discovery/blob/d498e7485b43fba711880cc73fdb916f49c8ccdd/lib/foreman_discovery/engine.rb#L147-L151) for example)

shlomizadok:
Thanks @xprazak2 !

PR #123, opened 2015-08-19
fixes #11424 - Add load_report.js to engine.rb assets:precompile

shlomizadok:


domcleal:
https://github.com/theforeman/foreman_plugin_template/blob/master/lib/foreman_plugin_template/engine.rb#L41-L55 is an alternative way of doing it for very simple cases (where all assets are top-level and won't require each other).  Or you could build a test around the technique.

shlomizadok:
@domcleal thank you so much for this! would love if you could ack.

domcleal:
It reads correctly, but I haven't tested that they are indeed precompiled.

shlomizadok:
@domcleal -- thanks. I've tested those. @xprazak2 / @ares - mind verifying?

xprazak2:
@shlomizadok: seems to be working for me, like [this](http://i.imgur.com/Y0Rv8Rt.png)

shlomizadok:
Thanks for testing @xprazak2 
Thanks for your insight @domcleal 
Merging

domcleal:
You're missing policy.css in that screenshot?

xprazak2:
hmm, maybe I am. Let me do a re-run.

shlomizadok:
the problem is that the file name is ```policy.css.scss``` and assets precompile expects ```policy.css```
As there is no real use of sass capabilities, I'm gonna rename that file in a separate pr.

PR #126, opened 2015-08-25
Fixes #11417 - Scaptimony migrated to foreman_openscap

xprazak2:
* models, migrations and seeds moved from scaptimony to foreman_openscap
* code from concerns moved to models, needless concerns deleted
* `Scaptimony::ArfReportHelper.create_arf` moved to `ForemanOpenscap::ArfReport.create_arf`
* removed dependency on scaptimony, added dependency on openscap

A couple of things on my mind:

1. Table names were changed from scaptimony_\* to foreman_openscap_\*. Since we do not have isolated namespace, we need to call `self.table_name="foreman_openscap_* " ` in each model. Furthermore, prefix is not omitted and forces changes in request params and even .js files. I find this kind of ugly. We can:

  A. keep it this way

  B. add `isolate_namespace`, which will restrict access to some of core's routing helpers and [break] (https://github.com/theforeman/foreman/blob/develop/app/views/host_mailer/_link_to_host.html.erb#L3) notification mailer (and maybe something else I do not know of yet...), therefore it would require additional changes. It may also be an inconvenience in the future views.

  C. remove the prefix entirely since there is no table name clash and we do not have isolated namespace in the first place. I like this option because it seems as the most straightforward one, but there may be a reason to keep the prefix that I am not aware of.

2. We do not depend on scaptimony any more, but there is still `Scaptimony::` namespace in services and many files have it in their names, which might get confusing. Do we want to keep it, remove it, or make a separate PR?

3. `Api::V2::Compliance::ArfReportsControllerTest` is now [broken](http://i.imgur.com/HHlwv6B.png). When I tweak it to show me what is wrong, it turns out, it cannot find arf_report_breakdowns table which may have something to do with its renaming and a fact that it is a view. I do not know how to convince the tests about its existence. Droping and recreating does not help, controller works as expected as far as I can tell.


app/models/concerns/foreman_openscap/model_extensions.rb
```diff
@@ -0,0 +1,8 @@
+module ForemanOpenscap
+  module ModelExtensions
+    extend ActiveSupport::Concern
+    included do
+      self.table_name_prefix = "foreman_openscap_"
```
shlomizadok: I think this should live in the engine. See [foreman_tasks engine](https://github.com/theforeman/foreman-tasks/blob/fd2206af4c9ce88ef08226ccac80a6d5224433ef/lib/foreman_tasks/engine.rb#L110-L112) for example

shlomizadok:
@xprazak2 - wow!
Regarding some of your concerns:
1. I have no problem with the table prefix and I'd like to keep it this way, yet ModelExtensions should be deleted and the table_name_prefix should move to engine.
2. ```Scaptimony::``` namespace and specifically controllers named Scaptimony, should be renamed. But let's do it in a different PR
3. Seems like tests are fixed?

xprazak2:
@shlomizadok: I made some changes:
1. The prefix is here to stay, I've added `table_name_prefix` into engine, I've also just found out there is `use_relative_model_naming?` method, so there will be no changes in params and .js files :smiley:

2. I've created an issue in Redmine

3. I have a sneaky suspicion openscap tests are not run, or at least I could not find them. But maybe I was looking in a wrong [place](http://ci.theforeman.org/job/test_plugin_pull_request/1217/database=sqlite3,label=fast,ruby=1.9.3/testReport/). Could you check pls?

shlomizadok:
[test]

shlomizadok:
this is the [passing test](http://ci.theforeman.org/job/test_plugin_pull_request/1217/)

xprazak2:
@shlomizadok: ok, I am sorry for raising a false alarm

app/mailers/foreman_openscap/policy_mailer.rb
```diff
@@ -6,7 +6,7 @@ def policy_summary(options = {})
       user = ::User.find(options[:user])
       @time = options[:time] || 1.day.ago
 
-      @policies = Scaptimony::Policy.all.reject {|policy| policy.assets.map(&:host).compact.empty?}
+      @policies = Policy.all.reject {|policy| policy.assets.map(&:host).compact.empty?}
```
shlomizadok: Should this be ```::ForemanOpenscap::Policy``` ?

xprazak2: @shlomizadok: This works because we are inside ForemanOpenscap module. I will change it to be more explicit, I've just find out I need to update data in taxable_taxonomies table, so I will be making a new commit anyway.

db/migrate/20150821100137_migrate_from_scaptimony.rb
```diff
@@ -0,0 +1,59 @@
...
+        LEFT OUTER JOIN
+          foreman_openscap_xccdf_rule_results rule
+          ON arf.id = rule.arf_report_id
+        LEFT OUTER JOIN foreman_openscap_xccdf_results result
+          ON rule.xccdf_result_id = result.id
+        GROUP BY arf.id;
+          SQL
+
+    TaxableTaxonomy.where(:taxable_type => ["Scaptimony::ArfReport", "Scaptimony::Policy", "Scaptimony::ScapContent"])
+      .map { |t| t.taxable_type.sub!(/^Scaptimony/, "ForemanOpenscap")}.map(&:save)
```
ares: maybe `save!` would reveal issues sooner

ares:
LGTM but I'd like to try the upgrade on my env as well

xprazak2:
Changed `save` to `save!` in migration

isimluk:
I am speachless.

db/migrate/20150821100137_migrate_from_scaptimony.rb
```diff
@@ -0,0 +1,61 @@
...
+        LEFT OUTER JOIN
+          foreman_openscap_xccdf_rule_results rule
+          ON arf.id = rule.arf_report_id
+        LEFT OUTER JOIN foreman_openscap_xccdf_results result
+          ON rule.xccdf_result_id = result.id
+        GROUP BY arf.id;
+          SQL
+
+    taxonomies = TaxableTaxonomy.where(:taxable_type => ["Scaptimony::ArfReport", "Scaptimony::Policy", "Scaptimony::ScapContent"])
+    taxonomies.map { |t| t.taxable_type.sub!(/^Scaptimony/, "ForemanOpenscap")}
```
ares: this does not work, you map the taxonomies but throw the result away, either use `map!` or `taxonomies = taxonomies.map {...}`, similarilly in down

ares:
It seems I'm not the only one surprised by this PR, could you put some light on why do we merge these gems? I only found this 9 days old issue http://projects.theforeman.org/issues/11417. @isimluk was the OpenSCAP service you've mentioned earlier using Scaptimony internally? Is there any other Scaptimony-based project currently or planned?

xprazak2:
@ares: I only know it is a part of Arf reports refactoring as written [here](https://docs.google.com/document/d/1VNpxPMvvqgbjO40io1_Vj5AkTDMjlmyJHxwhglP7v3E/edit#heading=h.2a5jebz370aj) (user stories on 2nd page). I fixed the migration, it should run smoothly now.

mpreisler:
This is a big surprise for me as well. SCAPtimony is a separate project, there are plans to use it in other areas than just Foreman. Bundling will cut you off patches in the upstream project! And it will cut us from your patches.

Can you explain why this is being done? I found http://projects.theforeman.org/issues/11417 but the issue doesn't even touch the reasons.

ares:
My understanding is that it was agreed that parsing of reports (in terms of Foreman reports) should be moved to smart proxy for several reasons. So would it make more sense to continue using scaptimony and just stop using it's report parsing capabilities when it's not needed? We would probably need to break it to more gems so it's dependencies are optional.

shlomizadok:
@isimluk @mpreisler and all - I am sorry, maybe I should have give more publicity to this issue.
I have created a [discussion](https://groups.google.com/forum/?hl=en#!topic/foreman-dev/x9zOw8NYx44) on foreman-dev group, so we can discuss it more openly.
Let's discuss it over there and reach conclusions which may be best for all of us.

ares:
After a short sync-up with @isimluk we agreed this is the right path. I tested the migration and it seems to work for me, so you've got my :+1: I suppose next step is migrating to Foreman reports STI and new status STI features while the parsing will be done in new gem being used by smart proxy.

shlomizadok:
Thanks @xprazak2 !

PR #127, opened 2015-09-09
Fixes #11744 - assign locations / organizations to default scap conte…

shlomizadok:
…nt seed


ares:
Thanks @shlomizadok, merging!

PR #128, opened 2015-09-09
Fixes #11576 - Scaptimony namespace removed

xprazak2:
I removed the references to Scaptimony from codebase.


app/helpers/report_dashboard_helper.rb
```diff
@@ -1,4 +1,4 @@
-module ScaptimonyReportDashboardHelper
+module ReportDashboardHelper
```
shlomizadok: Maybe rename this to `ArfReportDashboardHelper` so there won't be confusion with Foreman's `Report`

app/views/hosts/_host_status.html.erb
```diff
@@ -7,9 +7,9 @@
     <td><%= _('Compliance') %></td>
     <td>
       <% if host.arf_reports.search_for('failed > 0').blank? %>
-        <%= link_to(report_event_column('O', "label-success"), scaptimony_host_path(host.id)) %>
+        <%= link_to(report_event_column('O', "label-success"), host_path(host.id)) %>
```
shlomizadok: Should this be `compliance_host_path`?

xprazak2:
We probably have an [orpahned route](https://github.com/theforeman/foreman_openscap/blob/master/config/routes.rb#L13), even on gemified 0.4.3 this route is available and there is no index template.

app/helpers/compliance_hosts_helper.rb
```diff
@@ -0,0 +1,31 @@
+module ComplianceHostsHelper
+  
+  Colors = {
+    :passed  => '#89A54E',
+    :failed  => '#AA4643',
+    :othered => '#DB843D',
```
ares: seems like duplication of ArfReportDashboardHelper::Colors, during fixing, could you please rename Colors to COLORS?

db/migrate/20150827123826_remove_scaptimony_permissions.rb
```diff
@@ -0,0 +1,9 @@
+class RemoveScaptimonyPermissions < ActiveRecord::Migration
+  def up
+    Permission.where(:resource_type => ["Scaptimony::Policy", "Scaptimony::ScapContent"]).map(&:destroy)
```
ares: should this be migrated to new ones rather than just simply deleting them?

xprazak2:
I changed the name of Colors constant and removed the unnecessary duplicate. We now rename the permissions instead of removing and recreating them.

shlomizadok:
LGTM,
tested and works as expected.
@xprazak2, Colors constant has already added right?
@ares, seems mergeable - mind to take another look before we do?
Thanks.

db/migrate/20150827123826_remove_scaptimony_permissions.rb
```diff
@@ -0,0 +1,17 @@
+class RemoveScaptimonyPermissions < ActiveRecord::Migration
+  def up
+    permissions = Permission.where(:resource_type => ["Scaptimony::Policy", "Scaptimony::ScapContent"])
+    permissions.each do |p|
+      p.resource_type = p.resource_type.sub(/^Scaptimony/, "ForemanOpenscap")
```
ares: could you add more info to this, if there is ever an error while saving it would be hard to debug, so something like 
```say "Converting permission #{p.id} with name '#{p.name}' of type #{p.resource_type}"```
would be helpful

File: app/helpers/policy_dashboard_helper.rb [hunk 6]
```diff
@@ -8,7 +8,7 @@
 # along with this software; if not, see http://www.gnu.org/licenses/gpl.txt
 #
 
-module ScaptimonyPolicyDashboardHelper
+module PolicyDashboardHelper
   Colors = {
```
ares: I know it's completely unrelated but could I kindly ask you to fix also this one? It uses the same colors so you could either define the constant like this
```ruby
COLORS = {
  :compliant_hosts => ArfReportDashboardHelper::COLORS[:passed],
  :incompliant_hosts => ArfReportDashboardHelper::COLORS[:failed],
}
```

or use ArfReportDashboardHelper::COLORS in this helper directly.

ares:
Two nits, otherwise LGTM, I'll test the migration afterwards.

xprazak2:
I changed the COLOR constant in PolicyDashboardHelper as suggested and added debugging message to the migration

app/helpers/policy_dashboard_helper.rb
```diff
@@ -8,16 +8,16 @@
...
-    :compliant_hosts => '#89A54E',
-    :incompliant_hosts => '#AA4643',
-    :inconclusive_hosts => '#DB843D',
+module PolicyDashboardHelper
+  COLORS = {
+    :compliant_hosts => ArfReportDashboardHelper::COLORS[:passed],
+    :incompliant_hosts => ArfReportDashboardHelper::COLORS[:failed],
+    :inconclusive_hosts => ArfReportDashboardHelper::COLORS[:othered],
     :report_missing => '#92A8CD',
   }
```
shlomizadok: Excellent, however, this is not applied later in [line 37 ](https://github.com/xprazak2/foreman_openscap/blob/openscap-refactoring-backup-2/app/helpers/policy_dashboard_helper.rb#L37) and it raises `uninitialized constant PolicyDashboardHelper::Colors` error when visiting the policy's dashboard page.

app/overrides/hosts/overview/host_compliance_status.rb
```diff
@@ -1,4 +1,4 @@
 Deface::Override.new(:virtual_path  => "hosts/_overview",
                      :name          => "add_compliance_check",
                      :insert_after => "#properties_table",
-                     :partial       =>  "scaptimony_hosts/host_status")
\ No newline at end of file
+                     :partial       =>  "compliance_hosts/host_status")
```
shlomizadok: nitpick: add an extra line.

app/views/policies/create.html.erb
```diff
@@ -0,0 +1,2 @@
+<% title _("New Compliance Policy") %>
+<%= render :partial => "policies/steps/step_form" %>
```
shlomizadok: nitpick: add an extra line

app/views/policies/new.html.erb
```diff
@@ -0,0 +1,2 @@
+<% title _("New Compliance Policy") %>
+<%= render :partial => "policies/steps/step_form" %>
```
shlomizadok: nitpick: add an extra line

shlomizadok:
@xprazak2 - thanks for the effort here. I have found one issue, in the colors use in policy dashboard.
Other than that (and nitpicks) it looks very good. Thanks!

xprazak2:
@shlomizadok, policy dashboard should display without errors now, I also added the missing newlines.

shlomizadok:
Thanks @xprazak2 ! merging

PR #131, opened 2015-09-23
Fixes #11872 - Compliance status for hosts

xprazak2:
Current behaviour:
* When Arf report is uploaded to Foreman, Compliance status is updated/created for a host
* We have 3 compliance states: compliant, incompliant, inconclusive.
* When any latest arf report for any of the policies that are assigned to the host has a failed rule, the status is incompliant
* When latest arf reports for all policies assigned to the host have only passed rules, the status is compliant
* When the latest reports have passed and othered rules, the status is inconclusive

Changes in UI: 
* I removed the square icon with letter from host list
* I removed the Compliance table from host details, added a button to the [details](http://i.imgur.com/qI4i0qW.png) 

Questions and thoughts:
* Do we want to add Compliance status to API?
* We have one status for all policies on the host. If we wanted greater detail, we could create 'partial' scap status for 1 policy and then build 'global scap status' for a host. This would probably require changes in how Foreman displays statuses on host details page in properties table. A suitable place to display these 'partial statuses' could be a policy dashboard page


shlomizadok:
@xprazak2 - please rebase

xprazak2:
@shlomizadok: I rebased

ares:
Quick scan looks good, I'd like to give it another round and test later. Don't forget to either bump https://github.com/theforeman/foreman_openscap/blob/master/lib/foreman_openscap/engine.rb#L38 to 1.10 or wrap the status registration in a condition to keep 1.9- compatibility.

> Do we want to add Compliance status to API?
It should be added automatically, did you check the host detail output?

app/models/concerns/foreman_openscap/host_extensions.rb
```diff
@@ -8,6 +8,7 @@ module HostExtensions
       has_many :asset_policies, :through => :asset, :class_name => "::ForemanOpenscap::AssetPolicy"
       has_many :policies, :through => :asset_policies, :class_name => "::ForemanOpenscap::Policy"
       has_many :arf_reports, :through => :asset, :class_name => '::ForemanOpenscap::ArfReport'
+      has_one :scap_status, :class_name => 'HostStatus::ScapStatus', :foreign_key => 'host_id', :dependent => :destroy
```
ares: does `:dependent => :destroy` play nicely with host_statuses that also defines this dependency? (https://github.com/theforeman/foreman/blob/develop/app/models/host/managed.rb#L17) this has_one is subset of host_statuses

app/models/concerns/foreman_openscap/host_extensions.rb
```diff
@@ -34,6 +35,10 @@ def scap_status_changed?(policy)
       !last_reports.first.equal? last_reports.last
     end
 
+    def latest_report_for_policy(policy)
```
ares: nit: we usually use `last` instead of `latest` elsewhere

app/models/concerns/foreman_openscap/host_extensions.rb
xprazak2: yes, it would be probably better to remove it

xprazak2:
@ares, I removed `:dependent => :destroy`, changed `latest_report_for_policy` to `last_report_for_policy`, added a check into engine for compatibility, and made sure status gets picked up by [API](http://i.imgur.com/lNWFiKV.png). There may be something wrong in how the global status gets rendered by core, it has the correct "Error" label but the number code should be '2'. Could you please check if it is only local?

app/models/concerns/foreman_openscap/host_extensions.rb
```diff
@@ -8,6 +8,7 @@ module HostExtensions
       has_many :asset_policies, :through => :asset, :class_name => "::ForemanOpenscap::AssetPolicy"
       has_many :policies, :through => :asset_policies, :class_name => "::ForemanOpenscap::Policy"
       has_many :arf_reports, :through => :asset, :class_name => '::ForemanOpenscap::ArfReport'
+      has_one :compliance_status, :class_name => 'HostStatus::ComplianceStatus', :foreign_key => 'host_id'
```
ares: this should be named `compliace_status_object`, you override it by method with same name below https://github.com/theforeman/foreman_openscap/pull/131/files#diff-c08b2f64bbed0d7a15b0f7241af53d04R50

xprazak2:
@ares: I changed compliance_status to compliance_status_object

app/models/concerns/foreman_openscap/host_extensions.rb
```diff
@@ -8,6 +8,7 @@ module HostExtensions
       has_many :asset_policies, :through => :asset, :class_name => "::ForemanOpenscap::AssetPolicy"
       has_many :policies, :through => :asset_policies, :class_name => "::ForemanOpenscap::Policy"
       has_many :arf_reports, :through => :asset, :class_name => '::ForemanOpenscap::ArfReport'
+      has_one :compliance_status_object, :class_name => 'HostStatus::ComplianceStatus', :foreign_key => 'host_id'
```
ares: this association is usually used in scoped search, but I don't see any new definitions in the PR, could you add some? see https://github.com/theforeman/foreman_remote_execution/pull/42/files#diff-e3b1be04f41d99a66e199b0060b84e63R14 for an example

xprazak2:
@ares: I added scoped search on compliance status

app/models/host_status/compliance_status.rb
```diff
@@ -0,0 +1,46 @@
+module HostStatus
```
ares: this should be also under `ForemanOpenscap` module and hence in app/models/foreman_openscap/host_status (host status might not be needed, but leaving this up to you)

app/models/concerns/foreman_openscap/host_extensions.rb
ares: > @ares: I added scoped search on compliance status

next time pls write a comment to original one so we can see that it's addressed

app/models/host_status/compliance_status.rb
```diff
@@ -0,0 +1,46 @@
...
+        HostStatus::Global::OK
+      when INCONCLUSIVE
+        HostStatus::Global::WARN
+      else
+        HostStatus::Global::ERROR
+      end
+    end
+
+    def relevant?
+      !host.policies.empty?
```
ares: how about positive form `host.policies.present?`, I think it's more readable

app/models/host_status/compliance_status.rb
```diff
@@ -0,0 +1,46 @@
...
+    end
+
+    def to_status(options = {})
+      latest_reports = host.policies.map { |p| host.last_report_for_policy p }.flatten.partition { |report| report.failed? }
```
ares: `host.policies.map { |p| host.last_report_for_policy p }`

this should be possible to write using SQL, maybe we just need `has_many :arf_reports, :through => :policies` in host? I don't expect too many policies for a single host but still, it's n+1 problem

app/models/host_status/compliance_status.rb
```diff
@@ -0,0 +1,46 @@
...
+      return INCOMPLIANT unless latest_reports.first.empty?
```
ares: I'd find this more readable without using `partition`:

```ruby
latest_reports = host.policies.map { |p| host.last_report_for_policy p }.flatten # ideally replaced by SQL
failed = latest_reports.any?(&:failed?)
return INCOMPLIANT if failed.present?

othered = latest_reports.any?(&:othered?)
return INCONCLUSIVE if othered.present?

COMPLIANT
```

ares:
> There may be something wrong in how the global status gets rendered by core, it has the correct "Error" label but the number code should be '2'. Could you please check if it is only local?

This can happen in development if rails autoloader reloads the code. Restart Foreman and see if it helps, it seems to work for me. The reason is that we refer to class name in engine.rb in initializer (the way that Foreman plugin currently works) but we should do it in `config.to_prepare` block... Anyway this should be OK in production env.

app/models/concerns/foreman_openscap/host_extensions.rb
```diff
@@ -8,11 +8,16 @@ module HostExtensions
...
+      has_one :compliance_status_object, :class_name => 'ForemanOpenscap::::ComplianceStatus', :foreign_key => 'host_id'
 
       scoped_search :in => :policies, :on => :name, :complete_value => true, :rename => :'compliance_policy',
                     :only_explicit => true, :operators => ['= ', '!= '], :ext_method => :search_by_policy_name
       scoped_search :in => :policies, :on => :name, :complete_value => true, :rename => :'compliance_report_missing_for',
                     :only_explicit => true, :operators => ['= ', '!= '], :ext_method => :search_by_missing_arf
+      scoped_search :in => :compliance_status_object, :on => :status, :rename => :compliance_status,
+                    :complete_value => {:compliant => ForemanOpenscap::::ComplianceStatus::COMPLIANT,
+                                        :incompliant => ForemanOpenscap::::ComplianceStatus::INCOMPLIANT,
+                                        :inconclusive => ForemanOpenscap::::ComplianceStatus::INCONCLUSIVE}
```
ares:
```
/home/ares/Projekty/Zdrojaky/foreman_openscap/app/models/concerns/foreman_openscap/host_extensions.rb:18: syntax error, unexpected ::, expecting '(' (SyntaxError)
...ompliant => ForemanOpenscap::::ComplianceStatus::COMPLIANT,
...  
```

app/views/compliance_hosts/_compliance_status.erb
```diff
@@ -0,0 +1,6 @@
+<%= link_to(
+        _('Complinace'),
```
ares: s/Complinace/Compliance/

app/models/concerns/foreman_openscap/host_extensions.rb
```diff
@@ -8,11 +8,16 @@ module HostExtensions
       has_many :asset_policies, :through => :asset, :class_name => "::ForemanOpenscap::AssetPolicy"
       has_many :policies, :through => :asset_policies, :class_name => "::ForemanOpenscap::Policy"
       has_many :arf_reports, :through => :asset, :class_name => '::ForemanOpenscap::ArfReport'
+      has_one :compliance_status_object, :class_name => 'ForemanOpenscap::::ComplianceStatus', :foreign_key => 'host_id'
```
ares: another syntax error

shlomizadok:
@xprazak2 - Not sure if it's this PR or previous, yet my dashboard now looks like:
image

Also, rubocop is failing.

Thanks

shlomizadok:
Merging. Thanks @xprazak2 !

PR #134, opened 2015-10-17
move openscap related to proxy processing

shlomizadok:
This branch should work with https://github.com/theforeman/smart_proxy_openscap/pull/17

It sends all scap content to the proxy for:
* validation
* extract policies
* view policy guide

### DIFF
--- a/app/lib/proxy_api/available_proxy.rb
+++ b/app/lib/proxy_api/available_proxy.rb
@@ -1,0 +1,26 @@
+module ::ProxyAPI
+  class AvailableProxy
+
+    HTTP_ERRORS = [
+        EOFError,
+        Errno::ECONNRESET,
+        Errno::EINVAL,
+        Net::HTTPBadResponse,
+        Net::HTTPHeaderSyntaxError,
+        Net::ProtocolError,
+        Timeout::Error
+    ]
+
+    def initialize(args)
+      @features = ::ProxyAPI::Features.new(args).features
+    end
+
+    def available?
+      begin
+        return true if @features.include?('openscap')
+      rescue *HTTP_ERRORS
+        return false
+      end
+    end
+  end
+end
--- a/app/models/foreman_openscap/scap_content.rb
+++ b/app/models/foreman_openscap/scap_content.rb
@@ -1,27 +1,35 @@
 require 'digest/sha2'
-require 'openscap/ds/sds'
-require 'openscap/source'
-require 'openscap/xccdf/benchmark'
 
 module ForemanOpenscap
   class DataStreamValidator < ActiveModel::Validator
     def validate(scap_content)
       return unless scap_content.scap_file_changed?
 
-      allowed_type = 'SCAP Source Datastream'
-      if scap_content.source.try(:type) != allowed_type
-        scap_content.errors[:base] << _("Uploaded file is not #{allowed_type}.")
+      unless SmartProxy.with_features('Openscap').any?
+        scap_content.errors.add(:base, _('No Proxy with OpenScap features'))
+        return false
+      end
+
+      if scap_content.proxy_url.nil?
+        scap_content.errors.add(:base, _('No Available Proxy to validate SCAP content'))
         return false
       end
 
       begin
-        scap_content.source.validate!
-      rescue OpenSCAP::OpenSCAPError => e
-        scap_content.errors[:base] << e.message
+        api = ProxyAPI::Openscap.new(:url => scap_content.proxy_url)
+        errors = api.validate_scap_content(scap_content.scap_file)
+        if errors && errors['errors'].any?
+          errors['errors'].each {|error| scap_content.errors.add(:scap_file, _(error))}
+          return false
+        end
+      rescue *ProxyAPI::AvailableProxy::HTTP_ERRORS => e
+        scap_content.errors.add(:base, _('No available proxy to validate. Returned with error: %s') % e)
+        return false
       end
 
-      unless (scap_content.scap_content_profiles.map(&:profile_id) - scap_content.benchmark_profiles.profiles.keys).empty?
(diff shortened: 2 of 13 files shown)


app/models/foreman_openscap/scap_content.rb
```diff
@@ -1,27 +1,23 @@
...
 
 module ForemanOpenscap
   class DataStreamValidator < ActiveModel::Validator
     def validate(scap_content)
       return unless scap_content.scap_file_changed?
 
-      allowed_type = 'SCAP Source Datastream'
-      if scap_content.source.try(:type) != allowed_type
-        scap_content.errors[:base] << _("Uploaded file is not #{allowed_type}.")
+      api = ProxyAPI::Openscap.new(:url => SmartProxy.with_features('Openscap').first.url) rescue nil
```
ares: why `rescue nil?` if that's just because smart proxy doesn't have to exist, I'd prefer some if condition rather than using exceptions to control flow

shlomizadok: the `rescue nil` is to prevent a situation where there is no smart proxy with with Openscap feature.
Not sure it's the perfect solution. Any other ways to prevent error? I just want to ensure we've got a smart proxy with Openscap feature and return an error if not.

ares: so how about this?
```ruby
if (proxy = SmartProxy.with_features('Openscap').first)
  api = ProxyAPI::Openscap.new(:url => proxy.url)
else
  scap_content.errors.add(:base, _('No Proxy with OpenScap features.'))
  return false
end
```

shlomizadok: @ares - changed here and on policy view guide. thanks

app/models/foreman_openscap/scap_content.rb
```diff
@@ -1,27 +1,25 @@
...
+      else
+        scap_content.errors.add(:base, _('No Proxy with OpenScap features.'))
         return false
       end
 
-      begin
-        scap_content.source.validate!
-      rescue OpenSCAP::OpenSCAPError => e
-        scap_content.errors[:base] << e.message
+      errors = api.validate_scap_content(scap_content.scap_file)
```
ares: this can fail on timeout (and other network issues) so it should be wrapped, also in scenario where I have 2 proxies, first is failing (down) it can't use the second, would it make sense to iterate over all proxies with_features 'Openscap' and `break` with first success? sorry I noticed this during testing, it will need changing a bit also the `if (proxy = ...` that you already changed

shlomizadok: refactored

app/lib/proxy_api/available_proxy.rb
```diff
@@ -0,0 +1,28 @@
...
+        Net::ProtocolError,
+        Timeout::Error
+    ]
+
+    def initialize(args)
+      @url = args[:url]
+      super args
+    end
+
+    def available?
```
ares: we should migrate to smart_proxy_dynflow on smart proxy side so we could use its API to check proxy load, but seems good enough for now

app/lib/proxy_api/available_proxy.rb
```diff
@@ -0,0 +1,28 @@
...
+      begin
+        return @url if parse(get('/features')).include?('openscap')
+      rescue *HTTP_ERRORS
+        return nil
+      end
+      nil
```
ares: unreachable line?

app/lib/proxy_api/available_proxy.rb
ares: I'd expect this method to return true/false since it ends with `?`

app/models/foreman_openscap/scap_content.rb
```diff
@@ -71,36 +79,36 @@ def to_label
...
+      profiles
     end
 
-    private
-    def source_init
-      OpenSCAP.oscap_init
-      OpenSCAP::Source.new(:content => scap_file)
+    def proxy_url
+      proxies = SmartProxy.with_features('Openscap')
+      proxies.each_with_index do |proxy, index|
```
ares: untested but seems more readable to me

```ruby
def proxy_url
  @proxy_url ||= SmartProxy.with_features('Openscap').each do |proxy|
    proxy = ProxyAPI::AvailableProxy.new(:url => proxy.url)
    break proxy if proxy.available?
  end
end
```

app/models/foreman_openscap/scap_content.rb
```diff
@@ -1,27 +1,35 @@
...
 
-      allowed_type = 'SCAP Source Datastream'
-      if scap_content.source.try(:type) != allowed_type
-        scap_content.errors[:base] << _("Uploaded file is not #{allowed_type}.")
+      unless SmartProxy.with_features('Openscap').any?
+        scap_content.errors.add(:base, _('No Proxy with OpenScap features'))
+        return false
+      end
+
+      if !scap_content.proxy_url
```
ares: maybe `if scap_content.proxy_url.nil?` to avoid negation?

app/lib/proxy_api/available_proxy.rb
```diff
@@ -0,0 +1,28 @@
+module ::ProxyAPI
+  class AvailableProxy < ::ProxyAPI::Resource
```
ares: this shouldn't be API class but some service object, it should ideally just wrap `ProxyAPI::Features` which already has public method to get all [features](https://github.com/theforeman/foreman/blob/develop/lib/proxy_api/features.rb#L8).

app/lib/proxy_api/available_proxy.rb
shlomizadok: refactored

shlomizadok:
@ares, cleaned, squashed, smashed EOL, added tests. Ack when you can

ares:
Thanks @shlomizadok, merging!

PR #135, opened 2015-10-18
Fixes #11420 - ArfReport < Report

xprazak2:
This is a preliminary version, ~~some of the test suites are still broken.~~ existing tests should be passing now, I will add some more
* ArfReport < Report
* I took the liberty to relax some of the rubocop's rules (line/class length), even now they are more strict than in core
* fixed rubocop offences
* arf reports stored and displayed in the same way as Puppet reports
* db has a foreign key constraint on reports table that needs to be removed (something like [this](https://github.com/xprazak2/foreman/commit/c980fa7a50605f5820d59eb3e3be64feff36248d))


app/controllers/arf_reports_controller.rb
```diff
@@ -16,18 +16,16 @@ def index
...
   end
 
-  def parse
-    self.response_body = @arf_report.to_html
-  end
-
   def destroy
     if @arf_report.destroy
-      process_success
+      process_success(:success_msg => _(_("Successfully deleted Arf report.")), :success_redirect => arf_reports_path)
```
ares: double translations?

app/models/concerns/foreman_openscap/assetmix.rb
```diff
@@ -0,0 +1,7 @@
+module ForemanOpenscap
+  module Assetmix
```
ares: could you please rename to something more meaningful? e.g. BelongsToAssetMixin, also we tend to use ActiveSupport::Concern

app/models/foreman_openscap/arf_report.rb
```diff
@@ -1,112 +1,118 @@
...
-    has_one :arf_report_breakdown
     has_one :host, :through => :asset, :as => :assetable, :source => :assetable, :source_type => 'Host::Base'
 
     after_save :assign_locations_organizations
+    validate :result, :inclusion => { :in => RESULT }
 
     default_scope {
       with_taxonomy_scope do
-        order("foreman_openscap_arf_reports.created_at DESC")
+        order("reports.created_at DESC")
```
ares: please use `#{self.table_name}.created_at DESC` as we tend to rename tables from time to time

test/factories/policy_arf_report_factory.rb
```diff
@@ -0,0 +1,6 @@
+FactoryGirl.define do
+  factory :policy_arf_report, :class => ForemanOpenscap::PolicyArfReport do
+    policy_id nil
+    arf_report_id nil
+  end
+end
```
shlomizadok: Nit: EOL

test/factories/compliance_log_factory.rb
```diff
@@ -0,0 +1,11 @@
...
+  factory :compliance_log, :class => :log do
+    result "fail"
+    report
+    level_id 1
+    source nil
+    after(:build) do |log|
+      log.message = FactoryGirl.create(:message)
+    end
+  end
+end
```
ares: nit: EOL

test/factories/policy_arf_report_factory.rb
ares: nit: EOL

test/factories/compliance_log_factory.rb
ares: nit: EOL

test/unit/arf_report_status_calculator_test.rb
```diff
@@ -0,0 +1,11 @@
...
+
+class ArfReportStatusCalculatorTest < ActiveSupport::TestCase
+
+  test 'it should save metrics as bits' do
+    calc = ForemanOpenscap::ArfReportStatusCalculator.new(:counters => { 'passed' => 25, 'othered' => 1024, 'failed' => 10 })
+    assert_equal 25, calc.status['passed']
+    assert_equal ForemanOpenscap::ArfReport::MAX, calc.status['othered']
+    assert_equal 10, calc.status['failed']
+  end
+end
```
ares: nit: EOL

ares:
> db has a foreign key constraint on reports table that needs to be removed (something like this)

would you mind reporting it as core issue and sending a PR with the migration you provided as example?

app/models/foreman_openscap/arf_report.rb
```diff
@@ -130,10 +136,12 @@ def othered?
     end
 
     def equal?(other)
-      results = [xccdf_rule_results, other.xccdf_rule_results].flatten.group_by(&:xccdf_rule_id).values
+      results = [logs, other.logs].flatten.group_by(&:source_id).values
       # for each rule, there should be one result from both reports
       return false unless results.map(&:length).all? { |item| item == 2 }
-      results.map { |result| result.first.xccdf_result_id == result.last.xccdf_result_id }.all? && asset_id == other.asset_id && policy_id == other.policy_id
+      results.map { |result| result.first.source_id == result.last.source_id }.all? &&
```
ares: I know it's not your code originally, but the first condition part should be equivalent to
```ruby
results.all? { |result| result.first.source_id == result.last.source_id } &&
```

but more effective, since it stops on first mismatch

app/models/foreman_openscap/arf_report.rb
```diff
@@ -1,112 +1,118 @@
...
-              SELECT MAX(id) FROM foreman_openscap_arf_reports sub
-              WHERE sub.asset_id = foreman_openscap_arf_reports.asset_id)' }
-      when 'policy'
-        { :conditions => 'foreman_openscap_arf_reports.id IN (
-              SELECT MAX(id) FROM foreman_openscap_arf_reports sub
-              WHERE sub.policy_id = foreman_openscap_arf_reports.policy_id)' }
-      else
-        fail "Cannot search last by #{by}"
+    def rules_count
+      status.inject(0) { |memo, (k, v)| memo + v }
```
ares: slightly more readable `status.values.inject(&:+)`, or since we have activesupport, `status.values.sum` should work

app/models/foreman_openscap/arf_report.rb
```diff
@@ -1,112 +1,118 @@
...
+      # fail if policy does not exist.
+      policy = Policy.find(params[:policy_id])
+      ArfReport.transaction do
+        # TODO:RAILS-4.0: This should become arf_report = ArfReport.find_or_create_by! ...
+        arf_report = ArfReport.create!(:host_id => asset.id,
+                                       :reported_at => Time.at(params[:date].to_i),
+                                       :status => params[:metrics],
+                                       :metrics => params[:metrics])
+        PolicyArfReport.where(:arf_report_id => arf_report.id, :policy_id => policy.id, :digest => params[:digest]).first_or_create!
+        params[:logs].each do |log|
```
ares: if there's some issue with the report (e.g. no rule applicable or selected) then we receive no log, `params[:logs]` is nil which fails with exception here, not sure if we should treat that as failed report

ares:
Compliance status does not seem to be refreshed

app/views/arf_reports/_metrics.html.erb
```diff
@@ -0,0 +1,37 @@
...
+    </div>
+  </div>
+  <div class="col-md-4">
+    <div class="stats-well">
+      <h4 class="ca" ><%= _('Report Status') %></h4>
+      <%= flot_bar_chart("status" ,"", _("Number of Events"), status, :class => "statistics-bar")%>
+    </div>
+  </div>
+  <div class="col-md-4">
+    <table class='table table-bordered table-striped' style="height: 398px;">
```
ares: drop the height here, we have only three results

app/models/concerns/foreman_openscap/host_extensions.rb
```diff
@@ -45,9 +45,11 @@ def last_report_for_policy(policy)
 
     def reports_for_policy(policy, limit = nil)
       if limit
-        ForemanOpenscap::ArfReport.where(:asset_id => asset.id, :policy_id => policy.id).limit limit
+        ForemanOpenscap::ArfReport.joins(:policy_arf_report)
+          .merge(ForemanOpenscap::PolicyArfReport.of_policy policy.id).where(:host_id => asset.id).limit limit
```
shlomizadok: should be `asset.assetble_id`

app/models/concerns/foreman_openscap/host_extensions.rb
```diff
@@ -45,9 +45,11 @@ def last_report_for_policy(policy)
...
       else
-        ForemanOpenscap::ArfReport.where(:asset_id => asset.id, :policy_id => policy.id)
+        ForemanOpenscap::ArfReport.joins(:policy_arf_report)
+          .merge(ForemanOpenscap::PolicyArfReport.of_policy policy.id).where(:host_id => asset.id)
```
shlomizadok: ditto

app/models/foreman_openscap/arf_report.rb
```diff
@@ -1,112 +1,118 @@
...
+    def rules_count
+      status.values.sum
+    end
+
+    def self.create_arf(asset, params)
+      # fail if policy does not exist.
+      policy = Policy.find(params[:policy_id])
+      ArfReport.transaction do
+        # TODO:RAILS-4.0: This should become arf_report = ArfReport.find_or_create_by! ...
+        arf_report = ArfReport.create!(:host_id => asset.id,
```
shlomizadok: Should be `asset.assetable_id`

app/models/foreman_openscap/arf_report.rb
```diff
@@ -1,112 +1,118 @@
...
+                                       :reported_at => Time.at(params[:date].to_i),
+                                       :status => params[:metrics],
+                                       :metrics => params[:metrics])
+        PolicyArfReport.where(:arf_report_id => arf_report.id, :policy_id => policy.id, :digest => params[:digest]).first_or_create!
+        params[:logs].each do |log|
+          src = Source.find_or_create(log[:source])
+          #TODO: messages
```
ares: you can set message string to either `log[:title]` or maybe combine `log[:title] + log[:description] + log[:rationale]`, just note that there are "\n" characters that should be probably converted to <br> when displaying it

app/models/foreman_openscap/arf_report.rb
shlomizadok: `:asset_id => asset.id` should remain or no relationship between asset and report will be created

app/views/arf_reports/_list.html.erb
```diff
@@ -10,7 +10,7 @@
   <% for arf_report in @arf_reports %>
     <tr>
       <td><%= name_column(arf_report.host) %></td>
-      <td><%= _("%s ago") % time_ago_in_words(arf_report.date) %></td>
+      <td><%= _("%s ago") % time_ago_in_words(arf_report.reported_at) %></td>
       <td><%= report_arf_column(arf_report.passed, "label-info") %></th>
       <td><%= report_arf_column(arf_report.failed, "label-danger") %></th>
       <td><%= report_arf_column(arf_report.othered, "label-warning") %></th>
```
shlomizadok: in [line 21](https://github.com/theforeman/foreman_openscap/pull/135/files#diff-71e7815f8e3d6c41d1c1a504ae6266f4L21) should be `arf_report.host` instead of `arf_report.asset.host`

app/models/foreman_openscap/arf_report.rb
shlomizadok: I think Message should be `log[:title]`.
I'd either expand Message object or create a new model which includes: `description`, `rationale` & `fixes` as those SCAP components are very important to the users.

db/migrate/20150929152345_move_arf_reports_to_reports_table.rb
```diff
@@ -0,0 +1,179 @@
...
+        message = Message.find_or_create("No message for this log")
+
+        rule_item = xccdf_rule(rr_item['xccdf_rule_id'])
+        source = Source.find_or_create(rule_item['xid'])
+
+        Log.create!(:report_id => arf.id,
+                    :result => xccdf_result(rr_item["xccdf_result_id"])['name'],
+                    :message_id => message.id,
+                    :source_id => source.id,
+                    :level => :info)
```
shlomizadok: Should level be log[:severity]? So it would be "high / low / medium"? (Not sure..)

shlomizadok:
Some UI considerations:
Currently our puppet reports look like:
image
And the Arf Reports look like:
image

I think we need to unify the way they look as much as possible.
We are also missing a "view full report(html)" | "Download report(.xml.bzip2)" buttons to download from the proxy.

shlomizadok:
@xprazak2 - please rebase (sorry...)

app/models/foreman_openscap/arf_report.rb
ares: right, adding extra fields would work for me

bkearney:
+1 to the download raw report button/link. Is it possible to do the same level of charting on the puppet reports?

ares:
@bkearney download will be tricky as we don't know from which proxy we should download it, we can't even rely on authentication phase so we should probably somehow send this information with the report upload (but proxy does not know its identity in Foreman)

bkearney:
@ares TBH, I never understood the value of storing the data on the proxy. However, getting to the raw data (UI and CLI) will be important in order to qualify as a certifified scanning technology

shlomizadok:
@bkearney - Agreed. This will be added in a separate PR. Sounds good?

bkearney:
@shlomizadok sounds good

xprazak2:
The reports downloading from proxy should be working now, I pushed the changes as soon as I could to make them available, will add tests.

app/models/concerns/foreman_openscap/host_extensions.rb
xprazak2: @shlomizadok: It is correct as it is, host_id in arf_report points to asset, same goes for the following cases

db/migrate/20150929152345_move_arf_reports_to_reports_table.rb
xprazak2: @shlomizadok: we would have to define our own mapping onto existing [levels](https://github.com/shlomizadok/foreman/blob/fix_4151/app/models/log.rb#L9)
notice/warning/alert for low/medium/high perhaps?

shlomizadok: @xprazak2 - I am not sure. I feel that the "info" level is irrelevant in our use case. 
Either we drop it from the view, or map our severity to existing levels.
@ares - what do you think?

shlomizadok:
Note: tests are failing with: `LoadError: cannot load such file -- foreman_openscap/compliance_proxy`

app/assets/javascripts/foreman_openscap/load_report.js
```diff
@@ -2,5 +2,5 @@ $(document).ready(function() {
     $('#frame').load(function() {
         $('#loading').hide();
         $('#frame').show().css('min-height', '800px');
-    });    
-});
+    });
+});
```
ares: EOL

shlomizadok:
@xprazak2 - I still think it should be `assetable_id`
Here is my setting:
Host: `id`: `137`, `name`: `node10.local.lan`
Policy: `id`: `80`
Asset: `id`: `90`, `assetable_id`: `137`, `assetable_type`: `Host::Base`

Posting a new report raises the error:
```bash
"error": {"message":"ERROR:  insert or update on table \"reports\" violates foreign key constraint \"reports_host_id_fk\"\nDETAIL:  Key (host_id)=(90) is not present in table \"hosts\".\n"}
```

This is because it tries `Host.find(90)` in [foreman_openscap/app/models/foreman_openscap/arf_report.rb:104](https://github.com/theforeman/foreman_openscap/pull/135/files#diff-219a837adc650cf4c4ceae3c8a76eb7dR104)

xprazak2:
@shlomizadok: yes, I know about that - there is a a foreign key constraint on the reports table: `host_id references hosts`
There will be a change to make it work with constraint preserved, see [here](https://github.com/theforeman/foreman/pull/2857)

shlomizadok:
@xprazak2 why do we need to remove that constraint? We have that data and maybe we want to associate report with host?

xprazak2:
@shlomizadok: We would need to remove the constraint if we used host_id in reports to map to asset (current state). If we associate arf report with host directly, we can keep the constraint. That is what I am trying to do right now. Is this what you had in mind or did I misunderstood?

shlomizadok:
@xprazak2 - :+1:

shlomizadok:
@xprazak2 - tests are failing :cry: (also failed locally)
fails with:
```bash
Unknown column 'reports.managed_id' in 'where clause': SELECT COUNT(*) FROM `reports`  WHERE `reports`.`managed_id`
```

ares:
Last commit looks good to me, didn't have time to test though, @xprazak2 feel free to squash it. @shlomizadok if you tested it go ahead and merge.

xprazak2:
ok, I will squash

shlomizadok:
@xprazak2 - Thank you so much! Merging :100:

db/migrate/20151023131950_link_arf_report_directly_to_host.rb
```diff
@@ -0,0 +1,17 @@
+class LinkArfReportDirectlyToHost < ActiveRecord::Migration
+  def up
+    ForemanOpenscap::ArfReport.all.each do |report|
```
ohadlevy: I've tried this on my db - a few comments:

* please never try to loop on large tables such as reports, if you do, use find in batches instead.
* the logic here is odd, report already has a host_id, and then you are using that id to find the asset and save it again on the report host_id?
* third, please use update_all instead this will completely avoid the each loop.

xprazak2: I have submitted the improved version with batches.
The logic may seem strange, but there was a change from 
```ruby
class ArfReport
  belongs_to :asset, :foreign_key => 'host_id'
end

class Asset
  has_many :arf_reports
  belongs_to :assetable, :polymorphic => true
end

class Host::Managed
  has_one :asset, :as => :assetable
end
```
to
```ruby
class ArfReport
  belongs_to :host
end

class Asset
 belongs_to :assetable, :polymorphic => true
end

class Host::Managed
  has_many :arf_reports
  has_one :asset, :as => :assetable
end
``` 

so the real purpose is to replace asset ids in the 'host_id' column with the ids of the hosts.

PR #136, opened 2015-10-21
Fixes #12226 - don't fail if report is missing for policy

ares:


ares:
no_report

shlomizadok:
Thanks @ares!

PR #137, opened 2015-10-21
Fixes #12234 - display inherited policy status

ares:


shlomizadok:
@ares - Perfect. Thanks!

PR #138, opened 2015-11-03
refs #3157 - rails 4 changes

unorthodoxgeek:
related to https://github.com/theforeman/foreman/pull/2870

### DIFF
--- a/app/models/concerns/foreman_openscap/host_extensions.rb
+++ b/app/models/concerns/foreman_openscap/host_extensions.rb
@@ -55,6 +55,11 @@ module ForemanOpenscap
       assign_hostgroup_attributes %w(openscap_proxy_id)
     end
 
+    def policies=(policies)
+      self.create_asset(:assetable => self) if self.asset.blank?
+      self.asset.policies = policies
+    end
+
     def get_asset
       ForemanOpenscap::Asset.where(:assetable_type => 'Host::Base', :assetable_id => id).first_or_create!
     end
--- a/app/models/foreman_openscap/arf_report.rb
+++ b/app/models/foreman_openscap/arf_report.rb
@@ -15,7 +15,12 @@ module ForemanOpenscap
     has_one :policy, :through => :policy_arf_report
     has_one :asset, :through => :host, :class_name => 'ForemanOpenscap::Asset', :as => :assetable
     after_save :assign_locations_organizations
+    has_one :log, :foreign_key => :report_id
+
+    delegate :result, :to => :log, :allow_nil => true
     validate :result, :inclusion => { :in => RESULT }
+
+    delegate :asset=, :to => :host
 
     default_scope do
       with_taxonomy_scope do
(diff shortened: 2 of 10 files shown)


app/models/concerns/foreman_openscap/host_extensions.rb
```diff
@@ -46,6 +46,11 @@ module HostExtensions
       }
     end
 
+    def policies=(policies)
```
unorthodoxgeek: in rails 4, you can't set an association which goes through more than one has_many through, which policies does, thus, we do what rails 3 used to do for us, and create the asset if it doesn't exist yet, then set the policies

app/models/foreman_openscap/arf_report.rb
```diff
@@ -13,8 +13,13 @@ class ArfReport < ::Report
     has_one :policy, :through => :policy_arf_report
     has_one :asset, :through => :host, :class_name => 'ForemanOpenscap::Asset'
     after_save :assign_locations_organizations
+    has_one :log, :foreign_key => :report_id
+
+    delegate :result, :to => :log, :allow_nil => true
```
unorthodoxgeek: I have absolutely no clue how this worked until now, as there is no result field on the report. the validation should have thrown an exception in rails 3 as well...

db/migrate/20150112152944_create_scaptimony_arf_report_raws.rb
```diff
@@ -1,7 +1,7 @@
 class CreateScaptimonyArfReportRaws < ActiveRecord::Migration
   def change
     create_table :scaptimony_arf_report_raws, :id => false do |t|
-      t.references :arf_report, :index => true, :null => false
+      t.references :arf_report, :null => false
```
unorthodoxgeek: we create that index on line 8, no need for this (throws an exception because index already created)

app/models/foreman_openscap/arf_report.rb
xprazak2: Line 18 is a dead code and should be removed...

unorthodoxgeek: why?
there is no result method on ArfReport. as I said, I don't understand how this worked on rails 3, if anyone can explain this to me (and offer a working solution), I'd be more than happy to remove this code :smile:

without this line, ArfReports can't be created, as the validation tries to call #result, only to receive a NoMethodError

xprazak2: I apologize, I got confused and spoke too soon, you are right and I see it now. Please disregard my previous comment

unorthodoxgeek: I'm still VERY curious as to how this works in rails 3. what in the world answers to #result ? :open_mouth:

shlomizadok:
[test]

unorthodoxgeek:
shlomi! you're back!

this won't pass against current foreman develop... don't think the changes
are backward compatible

shlomizadok:
@unorthodoxgeek - thanks :)
Should this kept open until foreman develop is rails 4?

app/models/concerns/foreman_openscap/host_extensions.rb
```diff
@@ -47,6 +47,7 @@ module HostExtensions
       }
 
       alias_method_chain :set_hostgroup_defaults, :openscap
+      attr_accessible :discovery_rule_id
```
shlomizadok: Wrong plugin?

unorthodoxgeek: haha

shlomizadok:
@xprazak2 - can you test please?

xprazak2:
It worked for me without problems.

unorthodoxgeek:
@xprazak2 I can haz merge?

xprazak2:
@unorthodoxgeek: Yes! But I do not have permissions...

shlomizadok:
@unorthodoxgeek Thanks.
@xprazak2 Thanks for testing

PR #140, opened 2015-11-05
Fixes #12321 - Tests and Rubocop on PR should be running on Jenkins now

xprazak2:


shlomizadok:
So the tests are now triggering. However, they are failing. Would you mind fixing?
please note #138, which is supposed to fix some of those tests

xprazak2:
@shlomizadok: I will look into it

shlomizadok:
Perfect. Thanks @xprazak2 !

PR #141, opened 2015-11-18
Fixes #12524 - Add a migration to ensure mysql is creating a binary i…

shlomizadok:
…n the right size


ares:
LGTM

ShimiTaNaka:
If I'm not mistaking, this fix is incompatible with version 0.4.3 of the plugin, Can it be adapted?

shlomizadok:
@ShimiTaNaka - please see https://github.com/OpenSCAP/scaptimony/pull/32 for 0.4.3

ares:
0.4.* version used scaptimony for storing content (which seems to have the same [bug](https://github.com/OpenSCAP/scaptimony/blob/master/db/migrate/20141214112917_add_scap_file_to_scap_content.rb#L3))

PR #142, opened 2015-11-19
Fixes #9758 - Multiple Arf reports can be deleted simultaneously

xprazak2:
We can expire multiple Arf reports at the same time based on their created_at time. Reports older than supplied number of days will be expired. I am not sure if this is exactly what we want so I tried to keep it as simple as possible.

### DIFF
--- a/app/assets/javascripts/foreman_openscap/arf_reports.js
+++ b/app/assets/javascripts/foreman_openscap/arf_reports.js
@@ -1,0 +1,16 @@
+function buildArfModal(element, url) {
+  var url = url + "?" + $.param({arf_report_ids: $.foremanSelectedHosts});
+  var title = $(element).attr('data-dialog-title');
+  $('#confirmation-modal .modal-header h4').text(title);
+  $("#confirmation-modal .modal-body").load(url + " #content",
+      function(response, status, xhr) {
+        $('#submit_multiple').val('');
+        var b = $("#confirmation-modal .btn-primary");
+        if ($(response).find('#content form select').length > 0)
+          b.addClass("disabled").attr("disabled", true);
+        else
+          b.removeClass("disabled").attr("disabled", false);
+        $('#confirmation-modal').modal();
+        });
+  return false;
+}
--- a/app/controllers/arf_reports_controller.rb
+++ b/app/controllers/arf_reports_controller.rb
@@ -2,6 +2,7 @@ class ArfReportsController < ApplicationController
   include Foreman::Controller::AutoCompleteSearch
 
   before_filter :find_by_id, :only => [:show, :show_html, :destroy, :parse_html, :parse_bzip]
+  before_filter :find_multiple, :only => [:delete_multiple, :submit_delete_multiple]
 
   def model_of_controller
     ::ForemanOpenscap::ArfReport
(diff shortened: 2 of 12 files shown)


xprazak2:
Rubocop is failing, there already is PR making style changes, I expect this to turn green once #143 is merged

shlomizadok:
[test]

shlomizadok:
@xprazak2 - thanks for this PR. It works great :100: 
I have some UI concerns, though:
Maybe we should add a checkbox next to the reports and checking them will expire the reports?
We do so throughout Foreman (e.g., hosts#index). So one could select which reports to expire.
@ohadlevy @ares - what do you think?

app/controllers/arf_reports_controller.rb
```diff
@@ -45,6 +45,21 @@ def destroy
     end
   end
 
+  def expire_reports
+  end
+
+  def submit_expire_reports
+    days = Integer(params[:data][:days]) rescue nil
+    if days
+      ForemanOpenscap::ArfReport.expire :timerange => days.days
```
ohadlevy: won't this take a long time? we initially had an expire opt in the UI, and converted it into a rake / cron task instead.

app/views/arf_reports/expire_reports.html.erb
```diff
@@ -0,0 +1,23 @@
...
+
+<%= form_tag submit_expire_reports_arf_reports_path, :class => "form-horizontal well" do %>
+
+  <%= alert :class => 'alert-info', :header => _('Note'), :text => _('You are about to expire Compliance reports. All reports older than given number of day will be deleted. This action is irreversible') %>
+
+  <div class="clearfix">
+    <div class="form-group">
+      <%= label_tag("data[days]", _('Days'), :class => "col-md-2 control-label") %>
+      <div class="col-md-4">
+        <%= text_field(:data, :days, :placeholder => (_('Expire older than...')), :class => "form-control") %>
```
ohadlevy: why don't you use our bootstrap helpers instead of generating it yourself?

xprazak2: `text_f` and other our helpers are meant to be used inside block supplied to `form_for` helper. I use `form_tag` since `form_for` has to be bound to a model and I just need to get a numeric value.

xprazak2:
I will add checkboxes so we can select the reports directly.

app/views/arf_reports/expire_reports.html.erb
ohadlevy: you can still use the field helper afair

xprazak2:
I removed the expiring reports by time from UI and added a deletion based on checkboxes. I tried to reuse code from core as much as possible, but in some places it does not look very good.

app/views/arf_reports/delete_multiple.html.erb
```diff
@@ -0,0 +1,29 @@
...
+        <% end %>
+      </tbody>
+    </table>
+  </div>
+</div>
+
+<%= form_tag submit_delete_multiple_arf_reports_path({:arf_report_ids => params[:arf_report_ids]}) do %>
+  <span class="label label-danger"><%= _('Delete') %></span>
+  <%= _('these Complinace reports') %>
+<% end %>
```
shlomizadok: EOL

shlomizadok:
@xprazak2 - this looks very good. Can you please fix the rubocop errors?

app/controllers/arf_reports_controller.rb
```diff
@@ -45,16 +46,51 @@ def destroy
...
 
   def find_by_id
     @arf_report = resource_base.find(params[:id])
   end
 
+  def find_multiple
+    if params[:arf_report_ids].present?
+      @arf_reports = ::ForemanOpenscap::ArfReport.where(:id => params[:arf_report_ids])
+      if @arf_reports.empty?
+        error _('No Complince reports were found.')
```
shlomizadok: s/Complince/compliance

app/controllers/arf_reports_controller.rb
```diff
@@ -45,16 +46,51 @@ def destroy
...
+        redirect_to(arf_reports_path) and return false
+      end
+    else
+      error _('No Complince reports selected')
```
shlomizadok: ditto

app/assets/javascripts/foreman_openscap/arf_reports.js
```diff
@@ -0,0 +1,20 @@
+function buildArfModal(element, url) {
+  var url = url + "?" + $.param({arf_report_ids: $.foremanSelectedHosts});
+  console.log(url)
```
shlomizadok: remove

app/assets/javascripts/foreman_openscap/arf_reports.js
```diff
@@ -0,0 +1,20 @@
...
+  var title = $(element).attr('data-dialog-title');
+    $('#confirmation-modal .modal-header h4').text(title);
+    $('#confirmation-modal .modal-body').empty().append("<img class='modal-loading' src='/assets/spinner.gif'>");
+    $('#confirmation-modal').modal();
```
shlomizadok: First populate the modal body and then trigger it?

shlomizadok:
Merging. Thanks a lot @xprazak2 !

PR #144, opened 2015-11-30
Fixes #9766 - Openscap Proxy can be selected by a user

xprazak2:
* Dropdowns in host/hostgroup form that allow user to select openscap proxy for a host/hostgroup
* Each ArfReport is attached to the proxy and we are able to download report even if the host changes openscap proxy - as long as the proxy is registered in Foreman
* When the openscap proxy is destroyed, all associated reports are deleted from Foreman, but I do not bother doing cleanup (deleting reports from proxy) because we removed the proxy from Foreman and we do not care about it anymore.
* When host/hostgroup is saved with new openscap proxy, it configures the scap_client with the name and port from the new proxy.

### DIFF
--- a/app/controllers/concerns/foreman_openscap/hosts_common_controller_extensions.rb
+++ b/app/controllers/concerns/foreman_openscap/hosts_common_controller_extensions.rb
@@ -1,0 +1,45 @@
+module ForemanOpenscap
+  module HostsCommonControllerExtensions
+    extend ActiveSupport::Concern
+
+    def openscap_proxy_changed
+      model = if params[param_id_key].blank?
+                controller_name.classify.constantize.new(params[param_model_key])
+              else
+                resource_base.find(params[param_id_key])
+              end
+      proxy_id = params[param_model_key][:openscap_proxy_id]
+      instance_variable_set("@#{param_model_key}", model)
+      if proxy_id.blank?
+        render :partial => "form"
+      else
+        begin
+          instance_variable_get("@#{param_model_key}").update_scap_client_params(proxy_id)
+          render :partial => "form"
+        rescue => e
+          instance_variable_get("@#{param_model_key}").errors.add(:openscap_proxy_id, e.message)
+          render :partial => "form", :status => 422
+        end
+      end
+    end
+
+    def action_permission
+      case params[:action]
+      when 'openscap_proxy_changed'
+        :edit
+      else
+        super
+      end
+    end
+
+    private
+
+    def param_id_key
+      "#{param_model_key}_id"
+    end
+
+    def param_model_key
+      controller_name.singularize
+    end
+  end
+end
--- a/db/migrate/20151120090851_add_openscap_proxy_to_host_and_hostgroup.rb
+++ b/db/migrate/20151120090851_add_openscap_proxy_to_host_and_hostgroup.rb
@@ -1,0 +1,25 @@
+class AddOpenscapProxyToHostAndHostgroup < ActiveRecord::Migration
+  def up
+    add_column :hostgroups, :openscap_proxy_id, :integer
+    add_column :hosts, :openscap_proxy_id, :integer
+    add_column :reports, :openscap_proxy_id, :integer
+
+    #to ensure backward compatiblity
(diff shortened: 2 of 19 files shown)


xprazak2:
I got an idea to improve this, will add an update.

shlomizadok:
Please also rebase :)

app/controllers/concerns/foreman_openscap/hosts_common_controller_extensions.rb
```diff
@@ -0,0 +1,44 @@
+module ForemanOpenscap
+  module HostsCommonControllerExtensions
+    extend ActiveSupport::Concern
+
+    def openscap_proxy_changed
+      binding.pry
```
shlomizadok: Remove pleae :)

shlomizadok:
Tests failing because action was not added to permissions list. Please add in engine.rb

app/controllers/concerns/foreman_openscap/hosts_common_controller_extensions.rb
xprazak2: Oops!

app/controllers/concerns/foreman_openscap/hostgroups_controller_overrides.rb
```diff
@@ -0,0 +1,31 @@
+module ForemanOpenscap
+  module HostgroupsControllerOverrides
+    def self.included base
+      base.class_eval do
+        def load_vars_for_ajax
```
shlomizadok: We shouldn't replace core's `load_vars_for_ajax`, we should override it with `alias_method_chain` instead.
See [foreman_salt override](https://github.com/theforeman/foreman_salt/blob/master/app/controllers/foreman_salt/concerns/hostgroups_controller_extensions.rb#L24) for example.

also in `inherit_parent_attributes`

app/controllers/concerns/foreman_openscap/hosts_common_controller_extensions.rb
```diff
@@ -0,0 +1,44 @@
...
+
+    def param_id_key
+      "#{param_model_key}_id"
+    end
+
+    def param_model_key
+      controller_name.singularize
+    end
+  end
+end
```
shlomizadok: Nitpick: EOL

app/controllers/concerns/foreman_openscap/hosts_controller_overrides.rb
```diff
@@ -0,0 +1,35 @@
+module ForemanOpenscap
+  module HostsControllerOverrides
+    def self.included base
+      base.class_eval do
+        base::AJAX_REQUESTS << 'openscap_proxy_changed'
+
+        def process_hostgroup
```
shlomizadok: use `alias_method_chain`

app/views/compliance_hosts/_openscap_proxy.erb
```diff
@@ -0,0 +1,8 @@
+<%= javascript 'foreman_openscap/openscap_proxy' %>
```
shlomizadok: File name should be `_openscap_proxy.html.erb` ?

db/migrate/20151120090851_add_openscap_proxy_to_host_and_hostgroup.rb
```diff
@@ -0,0 +1,23 @@
...
+    add_column :hostgroups, :openscap_proxy_id, :integer
+    add_column :hosts, :openscap_proxy_id, :integer
+    add_column :reports, :openscap_proxy_id, :integer
+
+    #to ensure backward compatiblity
+    #this relies on the fact that only one scap proxy was registered
+    #because there has not been support for multiple scap proxies
+    reports = ForemanOpenscap::ArfReport.where(:openscap_proxy_id => nil)
+    scap_proxy = SmartProxy.with_features("Openscap").first
+    reports.each do |report|
```
shlomizadok: unless `scap_proxy.nil?`

shlomizadok:
Excellent. Merging. Thanks!

PR #147, opened 2015-12-20
Fixes #12885 - Remove forcing 'no whitespace' from policy name

shlomizadok:


shlomizadok:
Will fail on engine error.... :/

app/models/foreman_openscap/policy.rb
```diff
@@ -21,7 +21,7 @@ class Policy < ActiveRecord::Base
     SERVER_CLASS_PARAMETER   = 'server'
     PORT_CLASS_PARAMETER     = 'port'
 
-    validates :name, :presence => true, :uniqueness => true, :format => { :without => /\s/ }
+    validates :name, :presence => true, :uniqueness => true
```
ohadlevy: is there a reason why this existed in the first place? any limitation on scap scanner itself?

shlomizadok: Not as far as I know. And we use this name only to display it on foreman.
Maybe in the past it was used to set the puppet? We switched to id long time ago though

shlomizadok:
[test]

ares:
the test is failing because of this change, seems like it was on purpose

shlomizadok:
@ares, not sure [test] was unstable. (also tested locally and it worked)

ares:
let's see again but I think the [test] would have to be fixed too, IIRC it was no intermittent failure

xprazak2:
@shlomizadok could you rebase pls? I will set this as WoC for now...

shlomizadok:
@xprazak2 - rebase and fixed the test :)

xprazak2:
Works for me, we can merge if there are no objections.

ares:
:+1:

shlomizadok:
Thanks @ares & @xprazak2  for reviewing and testing

PR #148, opened 2016-01-12
Fixes #13135 - Policy mailer fix

xprazak2:
Policy mailer should work without problems now, I also added a test case that should catch these regressions in the future + one styling improvement


ares:
Tested, fixes the issue for me. Thanks @xprazak2, merging!

PR #152, opened 2016-03-06
Skip scap_content validation in rake db:seed if no scap proxy

shlomizadok:
Since 0.5.x we have moved scap validation to the proxy.
In rake db:seed, it is possible that smart_proxy_openscap feature is missing and
seed will fail.
As we are importing scap_security_guide content we can skip this validation if no proxy with
Openscap feature is present.

### DIFF
--- a/lib/foreman_openscap/bulk_upload.rb
+++ b/lib/foreman_openscap/bulk_upload.rb
@@ -15,10 +15,12 @@ module ForemanOpenscap
       end
 
       files_array = `rpm -ql scap-security-guide | grep ds.xml`.split
-      upload_from_files(files_array) unless files_array.empty?
+      # If no proxy with Openscap feature, we're gonna skip scap_security_guide files validation
+      skip_validation = (!SmartProxy.with_features('Openscap').any? && ensure_in_db_seed)
+      upload_from_files(files_array, skip_validation) unless files_array.empty?
     end
 
-    def upload_from_files(files_array)
+    def upload_from_files(files_array, skip_validation = false)
       files_array.each do |datastream|
         file = File.open(datastream, 'rb').read
         digest = Digest::SHA2.hexdigest(datastream)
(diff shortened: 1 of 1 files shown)


shlomizadok:
@ares, @isimluk would love your input on this

isimluk:
Please rebase on top of master. You are proposing changes, that have been already merged.

shlomizadok:
rebased... (oops)

ares:
Wouldn't it be safer to let user know that he must create a first proxy with scap plugin before he uploads content? Or there should be a way to validate it later when proxy exists. I'd lean towards the first, there's no point in including content if there's no scap proxy.

shlomizadok:
@ares - the change is happening only on the rake db:seed task (https://github.com/shlomizadok/foreman_openscap/blob/fix_14061/lib/foreman_openscap/bulk_upload.rb#L9) when uploading scap_security_guide (which we know is a valid scap content)

ares:
ah I see, still I'd expect https://github.com/theforeman/foreman_openscap/pull/152/files#diff-f1821d3c0696ad6b4d2024b406f5a900R19 to be checking `in_rake?` if in future someone will use bulk upload... anyway I do not insist, otherwise makes sense to me

shlomizadok:
@ares - Added. Thanks :smiley:

shlomizadok:
@ares - changed to a method which explicitly checks if in db:seed (as we have other rake tasks...)

lib/foreman_openscap/bulk_upload.rb
```diff
@@ -30,11 +32,19 @@ def upload_from_files(files_array)
         scap_content.original_filename = filename
         scap_content.location_ids = Location.all.map(&:id) if SETTINGS[:locations_enabled]
         scap_content.organization_ids = Organization.all.map(&:id) if SETTINGS[:organizations_enabled]
-        next puts "## SCAP content is invalid: #{scap_content.errors.full_messages.uniq.join(',')} ##" unless scap_content.valid?
-        if scap_content.save
-          puts "Saved #{datastream} as #{scap_content.title}"
+        if skip_validation
+          if (scap_content.save(:validate => false)) then
```
ares: please don't use then :-)

lib/foreman_openscap/bulk_upload.rb
```diff
@@ -30,11 +32,19 @@ def upload_from_files(files_array)
...
+            puts "Saved #{datastream} as #{scap_content.title}"
+          else
+            puts "Failed saving #{datastream}"
+          end
         else
-          puts "Failed saving #{datastream}"
+          next puts "## SCAP content is invalid: #{scap_content.errors.full_messages.uniq.join(',')} ##" unless scap_content.valid?
+          if scap_content.save then
```
ares: ditto

ares:
except one nit looks good to me, not tested

lib/foreman_openscap/bulk_upload.rb
shlomizadok: wtf??? (#godamnyourubymine)

shlomizadok:
Changed. This seems ready to merge (tested locally). Up to @ares

ares:
[test] after 5 tries, we got stable master build

PR #154, opened 2016-03-14
Remove db:seed job and introduce it as a rake task

shlomizadok:
1. Even if we skip validation, profiles will not be created and the scap content is irrelevant.
2. While seeding we can not ensure an available proxy.
3. Added to proxy availability its version as we need 0.5.x or higher.

### DIFF
--- a/app/lib/proxy_api/available_proxy.rb
+++ b/app/lib/proxy_api/available_proxy.rb
@@ -13,14 +13,27 @@ module ::ProxyAPI
 
     def initialize(args)
       @features = ::ProxyAPI::Features.new(args).features
+      @versions = ::ProxyAPI::Version.new(args).proxy_versions['modules']
     end
 
     def available?
       begin
-        return true if @features.include?('openscap')
+        return true if (@features.include?('openscap') && minimum_version)
       rescue *HTTP_ERRORS
         return false
       end
+      false
+    end
+
+    private
+
+    def openscap_proxy_version
+      @versions['openscap'] if @versions['openscap']
+    end
+
+    def minimum_version
+      return false unless openscap_proxy_version
+      openscap_proxy_version.to_f >= 0.5
     end
   end
 end
--- a/test/test_plugin_helper.rb
+++ b/test/test_plugin_helper.rb
@@ -14,6 +14,8 @@ Spork.each_run do
     def add_smart_proxy
       FactoryGirl.create(:smart_proxy, :url => 'http://localhost:8443', :features => [FactoryGirl.create(:feature, :name => 'Openscap')])
       ::ProxyAPI::Features.any_instance.stubs(:features).returns(%w(puppet openscap))
+      versions = { "version" => "1.11.0", "modules" => { "openscap" => "0.5.3" } }
+      ::ProxyAPI::Version.any_instance.stubs(:proxy_versions).returns(versions)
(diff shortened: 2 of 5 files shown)


ehelms:
Is foreman_openscap managed through Redmine? I notice sometimes changes get 'Fixes #' and sometimes they do not.

shlomizadok:
@ohadlevy, @ares - I had to revert the skip_validation as it created an issue with creating openscap_profiles (which are the core of why we need scap_content)
`rake foreman_openscap:bulk_upload:default` should create the scap-security-guide content instead.
It will raise errors if no proxy with openscap feature available or in the wrong version.

shlomizadok:
@ehelms http://projects.theforeman.org/projects/foreman_openscap/issues

ehelms:
Should we add this repository to this list? https://github.com/theforeman/prprocessor/blob/master/config/redmine_issue_required_repos.yaml

app/lib/proxy_api/available_proxy.rb
```diff
@@ -13,14 +13,26 @@ class AvailableProxy
...
+        return true if (@features.include?('openscap') && minimum_version)
       rescue *HTTP_ERRORS
         return false
       end
     end
+
+    private
+
+    def openscap_proxy_version
+      @versions['openscap'] if @versions['openscap']
```
ares: I think it can be simplified to `@versions['openscap']` unless you want to convert false -> nil, but I guess `@versions['openscap']` should never be false

ares:
except one nit, codewise :+1:, not sure why tests are failing, seems like you need to add some stubbing

shlomizadok:
[test]

app/lib/proxy_api/available_proxy.rb
shlomizadok: I'd like to convert false -> nil because nil.to_f = 0.0, false.to_f = error.
And @versions['openscap'] may be empty...

ares: if you mean false by empty then :+1: :-) I guess it doesn't hurt anything to be defensive anyway

shlomizadok:
@ares - Added stubbing. Tests pass. Thanks

test/test_plugin_helper.rb
```diff
@@ -14,6 +14,8 @@ class ActionController::TestCase
     def add_smart_proxy
       FactoryGirl.create(:smart_proxy, :url => 'http://localhost:8443', :features => [FactoryGirl.create(:feature, :name => 'Openscap')])
       ::ProxyAPI::Features.any_instance.stubs(:features).returns(%w(puppet openscap))
+      versions = { "version" =>"1.11.0", "modules" => { "openscap" => "0.5.3" } }
```
ares: nit: missing space before "1.11"
also I'm wondering if we can somehow stub it so it does not break after version bump

test/test_plugin_helper.rb
```diff
@@ -28,6 +30,8 @@ class ActiveSupport::TestCase
...
```
ares: ditto

test/test_plugin_helper.rb
shlomizadok: We actually don't care much about the Proxy's version. It is this way to formalize the response.
We are only checking ig openscap version is higher than 0.5 (which in turn is dependent on Proxy >= 1.11 on the gemspec...).
will fix the spacing

ares:
looks good to me

PR #157, opened 2016-04-04
Fixes #14446 - Respect STI class in taxonomy search

shlomizadok:
Not sure whether this should go here or in core.
@ares, @ohadlevy thoughts?
Anyhow tested w/ $downstream project and works as expected.

### DIFF
--- a/app/models/foreman_openscap/arf_report.rb
+++ b/app/models/foreman_openscap/arf_report.rb
@@ -173,6 +173,18 @@ module ForemanOpenscap
       end
     end
 
+    # can be removed after http://projects.theforeman.org/issues/14458 is fixed
+    def self.inner_select(taxonomy, inner_method = which_ancestry_method)
+      # always include ancestor_ids in inner select
+      conditions = { :taxable_type => self.base_class.name }
+      if taxonomy.present?
+        taxonomy_ids = get_taxonomy_ids(taxonomy, inner_method)
+        conditions.merge!(:taxonomy_id => taxonomy_ids)
+      end
+
+      TaxableTaxonomy.where(conditions).uniq.pluck(:taxable_id).compact
+    end
+
     def self.newline_to_space(string)
       string.gsub(/ *\n+/, " ")
     end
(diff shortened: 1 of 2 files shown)


ohadlevy:
I would open a refactoring issue to enable this usage case, and until its added, will add the workaround here.
However, if its a oneliner change (change from self.name to self.base_name) maybe just fix it in foreman itself? (at least raise a PR to start the discussion).

(in the larger scope - maybe you don't want to use taxonomix at all).

ohadlevy:
Also, I'm missing a test? :)

shlomizadok:
@ohadlevy not needed since https://github.com/theforeman/foreman/pull/3384 ?
image

shlomizadok:
@ohadlevy - Test has been added

ares:
LGTM, sounds as a good fix to core as well, if you open the redmine issue for it, please add a comment with link to this method so it can be tracked and removed later

shlomizadok:
Not so keen to add to core.
Will open an issue on redmine

ares:
> Not so keen to add to core.

why? isn't this broken for Compute Resources now? `Foreman::Model::Ovirt.name` and `Foreman::Model::Ovirt.base_class.name` returns different value

shlomizadok:
I'll re-phrase: Not so keen to add to core by myself.
Opened http://projects.theforeman.org/issues/14458, hopefully someone will pick it

app/models/foreman_openscap/arf_report.rb
```diff
@@ -173,6 +173,17 @@ def destroy
       end
     end
 
+    def self.inner_select(taxonomy, inner_method = which_ancestry_method)
```
ares: please add a comment "# can be removed after http://projects.theforeman.org/issues/14458 is fixed"

shlomizadok: Added

shlomizadok:
Closing in favor of https://github.com/theforeman/foreman/pull/3389 which is now merged and was asked to be ported into 1.11

PR #158, opened 2016-04-06
Fixes #14478 - Handling exception on search failure

xprazak2:
Turns out scoped search utterly fails with `NoMethodError` when there is a malformed query. Also `authorized?` helper did not work for me now for some reason (possibly change in the core).

### DIFF
--- a/app/controllers/arf_reports_controller.rb
+++ b/app/controllers/arf_reports_controller.rb
@@ -12,6 +12,8 @@ class ArfReportsController < ApplicationController
     @arf_reports = resource_base.includes(:host => [:policies, :last_report_object, :host_statuses])
       .search_for(params[:search], :order => params[:order])
       .paginate(:page => params[:page], :per_page => params[:per_page])
+  rescue NoMethodError
+    process_error :error_msg => _("Wrong search syntax")
   end
 
   def show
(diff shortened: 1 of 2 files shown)


shlomizadok:
This seems to be an issue with Core. try searching "pending" in config reports (http://localhost:3000/config_reports?utf8=%E2%9C%93&search=+pending+)

ares:
heh, that search kills my computer, the query took 77s and huge amount of RAM

shlomizadok:
I am sorry for the confusion. It seems to be a foreman_openscap issue after all.
I have removed the gem and the search mentioned above works.
In engine.rb we attach our compliance_status_scoped_search to Report model, hence it breaks it too
Regardless, I feel the proposed solution the right path to take.
Continuing to investigate.

ares:
Specifying just "pending" or any other string is not malformed query. It tries to search for %pending% on each scoped search definitions except those that have `:only_explicit => true`. It seems as some issue in scoped search, I thought it would be some `belongs_to` relation but we seem to have only `has_one` there.

shlomizadok:
@ares - true. But without foreman_openscap, when searching for "pending" does not raise an error :/

shlomizadok:
Please see: https://github.com/wvanbergen/scoped_search/issues/135
and: https://github.com/theforeman/foreman/pull/2466

shlomizadok:
@xprazak2 - thanks for the effort. Closing this in favor of https://github.com/theforeman/foreman/pull/3395 and will open a new pr that adds compliance_scoped_search only to ArfReport model.

PR #160, opened 2016-04-10
Fixes #14561 - Use 'patch' instead of 'put' in openscap_proxy.js

shlomizadok:
Otherwise, it yields 404 error, as 'patch' is sent instead of 'post'


ares:
It fixes the issue for editing host but when I try to create new one, I'm getting 404 since it tries to find host by id which does not exist yet.

shlomizadok:
@ares - Thanks for reviewing. I have remove ajax fetch for new hosts (don't think we update anything on the form, at this stage...) What do u think?

ares:
@xprazak2 will kindly look at it, I don't think it's correct that call `update_scap_client_params` from the form, it manipulates the DB... we should probably call it after form is submit (maybe before_save callback in host object)

xprazak2:
Unfortunately, changing the proxy should refresh the port and server parameters for foreman_scap_client puppetclass. Currently, `update_scap_client_params` does not work correctly for new host(group). Calling it `before_save` should help, but user will see old values in the UI. 
I will try to hide the parameters since we take care of them already and a user does not have to see what is going on behind the scene. Also my attempts to keep the form consistent failed and I am out of ideas :-(

ares:
so why this callback doesn't just build it as in memory objects without saving lookup values? that should render the form properly I guess

xprazak2:
Because the puppet params tab takes things from db when rendering the partial, see [here](https://github.com/theforeman/foreman/blob/develop/app/views/puppetclasses/_classes_parameters.html.erb#L12)

ares:
ah, that's sad :-/ hiding it is probably the best we can do now

shlomizadok:
I'm lost, what's the proposed course of action?

ares:
my understanding is that @xprazak2 works on a fix that will hide puppet parameters and they will be created/updated after form is submitted

shlomizadok:
OK, Still we need to /%s`put`/`patch` as Rails4 uses PATCH verb.

xprazak2:
I opened PR against Shlomi's branch. The LookupKeys/Values for foreman_scap_client should not be visible now at any point, no ajax is needed on changing openscap proxy. New proxy is set in callback when saving host(group)

shlomizadok:
@ares / @xprazak2 - PR is now updated

app/views/compliance_hosts/_openscap_proxy.html.erb
```diff
@@ -1,8 +1,6 @@
-<%= javascript 'foreman_openscap/openscap_proxy' %>
 <% data_url = @host ? openscap_proxy_changed_hosts_path : openscap_proxy_changed_hostgroups_path%>
 <%= select_f f, :openscap_proxy_id, SmartProxy.with_features("Openscap"), :id, :name,
     { :include_blank => blank_or_inherit_f(f, :openscap_proxy) },
     { :label => _('Openscap Proxy'),
-      :onchange => 'updateOpenscapProxy(this)',
       :'data-url' => data_url,
-      :help_inline => :indicator } %>
+      :help_inline => "" } %>
```
shlomizadok: Remove completely?

app/views/compliance_hosts/_openscap_proxy.html.erb
```diff
@@ -1,8 +1,6 @@
-<%= javascript 'foreman_openscap/openscap_proxy' %>
 <% data_url = @host ? openscap_proxy_changed_hosts_path : openscap_proxy_changed_hostgroups_path%>
 <%= select_f f, :openscap_proxy_id, SmartProxy.with_features("Openscap"), :id, :name,
     { :include_blank => blank_or_inherit_f(f, :openscap_proxy) },
     { :label => _('Openscap Proxy'),
-      :onchange => 'updateOpenscapProxy(this)',
       :'data-url' => data_url,
```
shlomizadok: Remove `data_url`?

xprazak2: Yeah, definitely. I will make a change

shlomizadok:
Rebased. @xprazak2 / @ares would love if you could take a look. Thanks

app/assets/javascripts/foreman_openscap/openscap_proxy.js
```diff
@@ -1,52 +0,0 @@
-function updateOpenscapProxy(element){
```
ares: I guess https://github.com/theforeman/foreman_openscap/blob/master/app/views/compliance_hosts/_openscap_proxy.html.erb#L1 should be removed as well

ares: as well as https://github.com/theforeman/foreman_openscap/blob/cd59dd809985b8bdd7bc9e4a304ae4da56cc25b7/app/views/compliance_hosts/_openscap_proxy.html.erb#L6

ares:
> New proxy is set in callback when saving host(group)

can you please point me to the callback? It does not seem to be triggered when updating host, the lookup value with proxy hostname is not updated. Also for new host I don't see any lookup keys being created.

xprazak2:
A bunch of stuff is missing now. [This](https://github.com/shlomizadok/foreman_openscap/pull/3/files) got merged in yesterday and none of it is here.

shlomizadok:
lost in rebasing? :/

shlomizadok:
Ooops... forgot to remove the "squash message" :smile: 
@xprazak2 - let me know if everything is in now.

xprazak2:
Seems like everything is in now.

ares:
Tests are failing

app/models/concerns/foreman_openscap/openscap_proxy_core_extensions.rb
```diff
@@ -4,18 +4,22 @@ module OpenscapProxyCoreExtensions
...
     included do
       validate :openscap_proxy_has_feature
+      validate :scap_client_class_present
+      after_save :update_scap_client
     end
 
-    def update_scap_client_params(proxy_id)
-      new_proxy = SmartProxy.find proxy_id
+    def update_scap_client
+      update_scap_client_params if openscap_proxy_id_changed? && openscap_proxy_id
```
ares: does this mean that it does not update values if I disassociate proxy? shouldn't it delete lookup values?

xprazak2: If you disassociate proxy, lookup values are not changed because you have puppetclasses for scap_client assigned. Note that user still has to assign classes to host manually (or make sure they are inherited from hostgroup). Ideally, user would just select/unselect openscap_proxy, and we would take care of assigning puppetclasses with correct lookup values. But the host form will display outdated/incorrect state of things with this approach. Therefore we have this "semi-manual" setup as a compromise.

ares: > But the host form will display outdated/incorrect state of things with this approach.

I though lookup values are hidden, so what would be outdated? When I disassociate the proxy I'd expect my foreman_scap_client manifest failing (setting wrong URL) unless I set this somewhere above in hierarchy. I'm about to test this first but it looks to me that just deleting lookup values if openscap_proxy_id is nil instead of updating would be better here.

xprazak2: >I though lookup values are hidden, so what would be outdated?

Lookup values were the greatest pain, but there is still foreman_scap_client puppetclass. If we want to add/remove it together with openscap_proxy, it will not be displayed correctly. And we cannot hide it since we add classes with ajax, but remove them purely client-side.
 
If you just delete lookup values for host, the Puppet will IMHO use defaults (haven't tested). If you disassociate openscap_proxy, why would you send reports to Foreman? Granted, you can still have a host with assigned policy that has a cron, but removing a proxy is clearly a sign that you no longer care about reports from this host. If you still care about the host, you will remove the foreman_scap_client manually, or you will delete the host completely. In both cases: no more reports. Scenario when someone removes the proxy and keeps the foreman_scap_client does not make much sense to me. 

Setting lookup values to something that will cause errors when disassociating proxy is probably a good idea, that way we can catch when someone disassociated proxy "by accident".

ares: > If we want to add/remove it together with openscap_proxy, it will not be displayed correctly.

I don't think we want. Selecting proxy does not say if the host should or should not run foreman_scap_client, it only tells which proxy is used to proxy the communication. It only affects lookup values in my eyes.

> If you just delete lookup values for host, the Puppet will IMHO use defaults

yes, which in this case won't be probably set anywhere which means `server` and `port` values will be missing in ENC output. That cause failure during puppet run since [puppet manifest](https://github.com/theforeman/puppet-foreman_scap_client/blob/master/manifests/init.pp#L38) does not define any default values. User in this case will see failed report which I think is correct.

> If you still care about the host, you will remove the foreman_scap_client manually, or you will delete the host completely. 

I don't think user would manually associate or disassociate this puppet manifest, they do it in policy wizard through host group and they therefore they can't remove it in host form. They can't even in host group form.

> Scenario when someone removes the proxy and keeps the foreman_scap_client does not make much sense to me.

If there's an advanced user who knows which lookup values must be set, he or she can have custom logic e.g. defining per host group, environment etc. Therefore they might want to disassociate the proxy later, in this case we'd keep overriding it on host level if we didn't delete lookup values.

> Setting lookup values to something that will cause errors when disassociating proxy is probably a good idea, that way we can catch when someone disassociated proxy "by accident".

Which is the case when you delete them. Unless advanced user set them on different level, which again would be correct.

PR #164, opened 2016-04-28
Resolves the problem with breakdown view on sqlite, fixes tests

xprazak2:
Another attempt to get rid of the errors. Seems like only sqlite had problems during migrations. I did not manage to reproduce locally, but I found out views in sqlite are read-only and cannot be altered. Now there should be no attempts to rename breakdown view and tests should turn green.


ares:
Fixed tests on my setup, thanks @xprazak2, merging!

PR #166, opened 2016-05-04
Fixes #14926 - Support for hammer plugin

xprazak2:


app/controllers/api/v2/compliance/arf_reports_controller.rb
```diff
@@ -31,7 +31,7 @@ def get_resource
         end
 
         resource_description do
-          resource_id 'foreman_openscap_arf_reports'
+          resource_id 'arf_reports'
```
ares: I'm not apipie expert, but based on documentation, this should not be needed anymore? https://github.com/Apipie/apipie-rails#resource-description

@shlomizadok do you happen to know what was the reason for custom resource_id?

shlomizadok: No, tbh.
I believe that all of the resource_description can be moved (from all of the controllers)

shlomizadok:
@xprazak2 - Thanks for this pull request. 
I suggest we'll remove the `resource_description` from our API controllers. I believe apipie will play nicely without them

xprazak2:
I removed `resource_description`, it works without it.

xprazak2:
I also changed `scap_content/show` in api controller so we could have the `info` command in hammer. The original was renamed to `xml`.

There were failures in `AvailableProxy` related code when proxy was down/unavailable, now we should get nice message in all cases.

I added a few more nodes into rabl views as well.

shlomizadok:
@xprazak2 - Thank you so much for this PR. Merging

PR #171, opened 2016-06-10
Fixes #15365 - Delete reports when cannot find proxy

xprazak2:
When proxy associated with reports is not available (or was already removed), the reports cannot be deleted because the request to delete a report file on proxy fails. This changes behaviour to 'best effort' - if proxy can be reached, try to do a cleanup, if not, just delete the report from db.

### DIFF
--- a/app/models/concerns/foreman_openscap/openscap_proxy_extensions.rb
+++ b/app/models/concerns/foreman_openscap/openscap_proxy_extensions.rb
@@ -10,7 +10,7 @@ module ForemanOpenscap
     def openscap_proxy_api
       return @openscap_api if @openscap_api
       proxy_url = openscap_proxy.url if openscap_proxy
-      fail(_("No OpenSCAP proxy found for %s") % id) unless proxy_url
+      raise ::Foreman::Exception.new(N_("No OpenSCAP proxy found for %{class} with %{id}"), { :class => self.class, :id => id }) unless proxy_url
       @openscap_api = ::ProxyAPI::Openscap.new(:url => proxy_url)
     end
   end
--- a/app/models/foreman_openscap/arf_report.rb
+++ b/app/models/foreman_openscap/arf_report.rb
@@ -164,11 +164,13 @@ module ForemanOpenscap
     end
 
     def destroy
-      if openscap_proxy_api.destroy_report(self, ForemanOpenscap::Helper::find_name_or_uuid_by_host(host))
-        super
-      else
-        false
+      begin
+        openscap_proxy_api.destroy_report(self, ForemanOpenscap::Helper::find_name_or_uuid_by_host(host))
+      rescue Foreman::Exception => e
+        logger.error "Failed to delete report with id #{id} from proxy, cause: #{e.message}"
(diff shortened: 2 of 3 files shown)


ohadlevy:
@shlomizadok can you review please?

app/models/foreman_openscap/arf_report.rb
```diff
@@ -164,11 +164,12 @@ def equal?(other)
...
 
     def destroy
-      if openscap_proxy_api.destroy_report(self, ForemanOpenscap::Helper::find_name_or_uuid_by_host(host))
-        super
-      else
-        false
+      begin
+        openscap_proxy_api.destroy_report(self, ForemanOpenscap::Helper::find_name_or_uuid_by_host(host))
+      rescue => e
+        logger.debug("Failed to delete report with id #{id} from proxy, cause: #{e.message}")
```
ohadlevy: why debug? shouldnt the user be aware it didnt work?

app/models/concerns/foreman_openscap/openscap_proxy_extensions.rb
```diff
@@ -10,7 +10,7 @@ module OpenscapProxyExtensions
     def openscap_proxy_api
       return @openscap_api if @openscap_api
       proxy_url = openscap_proxy.url if openscap_proxy
-      fail(_("No OpenSCAP proxy found for %s") % id) unless proxy_url
+      fail(_("No OpenSCAP proxy found for %{klass} with %{id}") % {:klass => self.class, :id => id}) unless proxy_url
```
shlomizadok: Nit: this can be `:class` ... :smile:

app/models/foreman_openscap/arf_report.rb
```diff
@@ -164,11 +164,12 @@ def equal?(other)
     end
 
     def destroy
-      if openscap_proxy_api.destroy_report(self, ForemanOpenscap::Helper::find_name_or_uuid_by_host(host))
-        super
-      else
-        false
+      begin
+        openscap_proxy_api.destroy_report(self, ForemanOpenscap::Helper::find_name_or_uuid_by_host(host))
+      rescue => e
```
shlomizadok: Should this rescue be more specific or are we permissive with the errors we're rescuing here?

shlomizadok:
@xprazak2 - Thanks. Got two minor issues. Can you please fix?

xprazak2:
We are now more selective in what exceptions we rescue, I also improved logging a bit. I tried to find a solution how to propagate the message to user as @ohadlevy suggested, but I did not like the result very much. It can be found [here](https://github.com/xprazak2/foreman_openscap/commit/9cfcc90a529d4f2735d539776addd92a01c542e3)

ohadlevy:
@shlomizadok ping?

shlomizadok:
@xprazak2 - Looks good. I am missing a test which covers this scenario. Mind to add ? Thanks

xprazak2:
I added tests to cover the change.

shlomizadok:
@xprazak2 - perfect. Thanks a lot! (Merging)

PR #173, opened 2016-06-21
Fixes #15477 - Change apipie doc for policy

xprazak2:
to make hammer_cli_foreman_openscap work properly when
creating new policy

### DIFF
--- a/app/models/foreman_openscap/policy.rb
+++ b/app/models/foreman_openscap/policy.rb
@@ -67,8 +67,8 @@ module ForemanOpenscap
 
     def hostgroup_ids=(ids)
       hostgroup_assets = []
-      ids.reject(&:empty?).map do |id|
-        hostgroup_assets << assets.where(:assetable_type => 'Hostgroup', :assetable_id => id).first_or_create!
+      ids.reject { |id| id.respond_to?(:empty?) && id.empty? }.map do |id|
+        hostgroup_assets << assets.where(:assetable_type => 'Hostgroup', :assetable_id => id).first_or_initialize
       end
       self.assets = hostgroup_assets
     end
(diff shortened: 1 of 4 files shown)


app/models/foreman_openscap/policy.rb
```diff
@@ -248,14 +249,18 @@ def ensure_period_specification_present
...
-        puppetclass = Puppetclass.find_by_name(SCAP_PUPPET_CLASS)
+        puppetclass = find_scap_puppetclass
         hostgroups.each do |hostgroup|
           hostgroup.puppetclasses << puppetclass unless hostgroup.puppetclasses.include? puppetclass
           populate_overrides(puppetclass, hostgroup)
         end
       end
     end
 
+    def find_scap_puppetclass
```
shlomizadok: It seems that this method is used only once. so not sure why to create it (unless plans to use more of it?)

app/models/foreman_openscap/policy.rb
```diff
@@ -67,8 +67,9 @@ def hostgroup_ids
 
     def hostgroup_ids=(ids)
       hostgroup_assets = []
-      ids.reject(&:empty?).map do |id|
-        hostgroup_assets << assets.where(:assetable_type => 'Hostgroup', :assetable_id => id).first_or_create!
+      ids.reject { |id| id.respond_to?(:empty?) && id.empty? }.map do |id|
+        params = { :assetable_type => 'Hostgroup', :assetable_id => id }
+        hostgroup_assets << (assets.where(params).first || Asset.new(params))
```
shlomizadok: Why not `.first_or_create` ?

xprazak2: It is called from policy.save and raises error:

app/models/foreman_openscap/policy.rb
xprazak2: I need to stub it in test because we check if foreman_scap_client is present.

app/models/foreman_openscap/policy.rb
shlomizadok: Cool. Alas, it still bothers me a bit :)
Would you mind testing it with `.first_or_initialize` ? (sorry for the hustle)

xprazak2: No problem, `first_or_initialize` seems to work.

shlomizadok: :+1:  Thanks. It's a better practice I believe

shlomizadok:
@xprazak2 -- looking forward to `hammer openscap policy --id=2` :smile: 
Merging. Thanks !

PR #177, opened 2016-07-12
Fixes #15654 - Remove deprecated initializer

xprazak2:


shlomizadok:
:+1: pending jenkins

xprazak2:
I do not think we will get stable build on Jenkins, it fails on what this PR is trying to solve...

PR #178, opened 2016-07-12
Fixes #15676 - defensive migration 

shlomizadok:
During migration, before compliance turns to foreman_openscap and later migrations
we encounter failures due to not finding 'openscap_proxy_id' and foreman_openscap_assets table.
this code is defending from those failures

### DIFF
--- a/app/models/concerns/foreman_openscap/openscap_proxy_core_extensions.rb
+++ b/app/models/concerns/foreman_openscap/openscap_proxy_core_extensions.rb
@@ -9,7 +9,7 @@ module ForemanOpenscap
     end
 
     def update_scap_client
-      update_scap_client_params if openscap_proxy_id_changed?
+      update_scap_client_params if changed.include?('openscap_proxy_id')
     end
 
     def update_scap_client_params
--- a/app/models/foreman_openscap/compliance_status.rb
+++ b/app/models/foreman_openscap/compliance_status.rb
@@ -37,6 +37,8 @@ module ForemanOpenscap
     end
 
     def relevant?
+      # May fail host status during migration
+      return false unless ForemanOpenscap::Asset.table_exists?
       host.policies.present?
     end
 


shlomizadok:
cc @tbrisker

ohadlevy:
@shlomizadok what is source of the problem? shouldnt it be a similar fix to https://github.com/theforeman/foreman/pull/3639 in core?

app/models/concerns/foreman_openscap/openscap_proxy_core_extensions.rb
```diff
@@ -9,7 +9,7 @@ module OpenscapProxyCoreExtensions
     end
 
     def update_scap_client
-      update_scap_client_params if openscap_proxy_id_changed?
+      update_scap_client_params if changed.include?('openscap_proxy_id')
```
ohadlevy: should we also check it includes the scap feature?

shlomizadok: No. This runs on host(group)

shlomizadok:
@ohadlevy - the issue here is that previously we had `Compliance` namespace and now it is `ForemanOpenSCAP` namespace - but Foreman migrations are running before foreman_openscap migrations

ohadlevy:
@tbrisker opinions? I guess I'm OK with this.

tbrisker:
I didn't test it (other then when @shlomizadok was writing it and showed me), but this looks good in principle. I think you should however use a separate issue number then the core PR, because of the automatic release process.

shlomizadok:
@ohadlevy @tbrisker - reworded. ACK if good to merge

tbrisker:
APJ as far as I can tell.

ohadlevy:
APJ

shlomizadok:
Merging. Thanks a lot @tbrisker & @ohadlevy

PR #179, opened 2016-07-13
Fixes #15476 - Improve validations for policy

xprazak2:


shlomizadok:
@xprazak2 will you please rebase? Thanks

app/models/foreman_openscap/policy.rb
```diff
@@ -168,8 +164,38 @@ def to_enc
       }.merge(period_enc)
     end
 
+    def should_validate?(step_i)
+      if new_record? && wizard_initiated?
+        step_index > step_i
+      elsif new_record? && !wizard_initiated?
+        true
+      else
+        !id.blank?
```
ares: should be equal to `persisted?` I think

app/models/foreman_openscap/policy.rb
```diff
@@ -168,8 +164,38 @@ def to_enc
...
+      end
+    end
+
+    def wizard_initiated?
+      @wizard_initiated
+    end
+
+    def update_attributes(hash)
```
ares: I'd recommend moving this to before_validation callback method for following reasons:
* this overrides internal #updates_attributes AR method, if they later change the signature of this method it will break
* if we reset in before_validation we get closer to point when the record is normally validates

ares: also validation would work on self instead of hash, `erase_period_attrs` could the look like this

```ruby
def erase_period_attrs(attrs)
  attrs.each { |attr| self.public_send("#{attr}=", nil) }
end
```

app/models/foreman_openscap/policy.rb
```diff
@@ -219,15 +245,30 @@ def ensure_needed_puppetclasses
...
-      cron_line.split(' ')
+      cron_line.to_s.split(' ')
     end
 
     def valid_cron_line
-      return true if period != 'custom' || step_index != 4
+      if period == 'custom' && should_validate?(4)
+        errors.add(:cron_line, _("does not consist of 5 parts separated by space")) unless cron_line_split.size == 5
+      else
+        errors.add(:cron_line, _("should be empty")) unless cron_line.blank?
```
ares: I think this could be implemented by standard `validate :cron_line, :presence => true, :if => :cron_line.present?`

app/models/foreman_openscap/policy.rb
```diff
@@ -21,14 +21,10 @@ class Policy < ActiveRecord::Base
...
     validate :ensure_needed_puppetclasses
     validates :period, :inclusion => {:in => %w(weekly monthly custom)},
-              :if                 => Proc.new { |policy| policy.new_record? ? policy.step_index > 3 : !policy.id.blank? }
-    validates :weekday, :inclusion => {:in => Date::DAYNAMES.map(&:downcase)},
-              :if                  => Proc.new { |policy| policy.period == 'weekly' && (policy.new_record? ? policy.step_index > 3 : !policy.id.blank?) }
-    validates :day_of_month, :numericality => {:greater_than => 0, :less_than => 32},
-              :if                          => Proc.new { |policy| policy.period == 'monthly'&& (policy.new_record? ? policy.step_index > 3 : !policy.id.blank?) }
-    validate :valid_cron_line
-    validate :ensure_period_specification_present
+              :if => Proc.new { |policy| policy.should_validate?(3) }
```
ares: could you please replace all step indexes in this PR with something like `steps.index('Schedule')`

app/models/foreman_openscap/policy.rb
xprazak2: I believe it would be more like:
`validate :cron_line, :presence => true :if => period == 'custom' && should_validate?(4)`
`validate :cron_line, :if => cron_line.present?, :with => something_that_checks_cron_line_format`
`validate :cron_line, :absence => true, :if => period != 'custom' && should_validate?(4)`

Because I think you want to check that cron line is present and has correct format `if period == 'custom'` and it should not be present at all `unless period == 'custom'`. And you have to keep track of the wizard step as well.

shlomizadok:
@xprazak2 - please note that tests are failing :/

app/models/foreman_openscap/policy.rb
```diff
@@ -168,8 +170,37 @@ def to_enc
...
+      else
+        persisted?
+      end
+    end
+
+    def wizard_initiated?
+      @wizard_initiated
+    end
+
+    def update_attrs
```
shlomizadok: `update_attrs` strikes me as too general name (e.g., I expected to find more attributes to update here) where this update deals only with `period`. Would you consider renaming this method?

shlomizadok:
@xprazak2 - Thank you for this improvement. Tested and it works well. I have only one nitpick regarding a method name (let me know what you thing - I may be wrong...)
:+1:

xprazak2:
I renamed the method to `update_period_attrs`. You are right @shlomizadok, it deals only with period attributes and the name was a bit misleading.

shlomizadok:
Merging, Thanks a lot @xprazak2 !

PR #181, opened 2016-07-20
Fixes #15756 - Respect host name/uuid

shlomizadok:
In migration, we should also respect the
host's uuid (with Katello) or name (without)

### DIFF
--- a/lib/foreman_openscap/helper.rb
+++ b/lib/foreman_openscap/helper.rb
@@ -16,25 +16,21 @@ module ForemanOpenscap::Helper
   end
 
   def self.find_name_or_uuid_by_host(host)
-    if defined?(::Katello::System)
-      host.content_host.uuid
-    else
-      host.name
-    end
+    (host.respond_to?(:subscription_facet) && !host.subscription_facet.nil?) ? host.subscription_facet.try(:uuid) : host.name
   end
 
   private
 
   def self.find_host_by_name_or_uuid(cname)
-    if defined?(Katello::System)
-      host = Host.includes(:content_host).where(:katello_systems => {:uuid => cname}).first
+    if Facets.registered_facets.keys.include?(:subscription_facet)
+      host = Katello::Host::SubscriptionFacet.find_by_uuid(cname).try(:host)
       host ||= Host.find_by_name(cname)
     else
       host = Host.find_by_name(cname)
     end
+
     unless host
       Rails.logger.error "Could not find Host with name: #{cname}"
-      Rails.logger.error "Please check that Content host is linked to Foreman host" if defined?(Katello::System)
       fail ActiveRecord::RecordNotFound
     end
     host
--- a/lib/foreman_openscap/data_migration.rb
+++ b/lib/foreman_openscap/data_migration.rb
@@ -64,7 +64,7 @@ module ForemanOpenscap
 
     def fetch_host_name(asset_id)
       asset = ForemanOpenscap::Asset.find(asset_id)
-      asset.host.to_label
+      ForemanOpenscap::Helper.find_name_or_uuid_by_host(asset.host)
     end
 
     def arfs_by_reported(time)


ehelms:
My take on this issue based on running upgrade myself and trying to view reports -- https://github.com/theforeman/foreman_openscap/pull/182

lib/foreman_openscap/helper.rb
```diff
@@ -16,25 +16,21 @@ def self.get_asset(cname, policy_id)
...
-    end
+    host.respond_to?(:subscription_facet) ? host.subscription_facet.uuid : host.name
   end
 
   private
 
   def self.find_host_by_name_or_uuid(cname)
-    if defined?(Katello::System)
-      host = Host.includes(:content_host).where(:katello_systems => {:uuid => cname}).first
+    if defined?(Katello::Host::SubscriptionFacet)
```
ehelms: I think `host.respond_to?` would be better here similar to line 19

shlomizadok: Yeah, but here we do not have the host object. We have the `cname` string

ehelms: Good point - what about using the Facets API?

```
Facets.registered_facets.keys.include?(:subscription_facet)
```

shlomizadok:
Tested with migration of Satellite 6.1 to 6.2 seems to work okay

lib/foreman_openscap/helper.rb
shlomizadok: Good point. Which is better?

ehelms: I would think the facets API should be the 'best' way to check for the existence of a facet rather than relying on a class that could change.

shlomizadok: Changed. Thanks @ehelms

ehelms:
I encountered an error here, the host will respond to this function but that does not mean it has a subscription_facet. To get around this issue when running the migrate rake task I had to modify this to:

```
module ForemanOpenscap::Helper
  def self.get_asset(cname, policy_id)
    asset = find_host_by_name_or_uuid(cname).get_asset
    asset.policy_ids += [policy_id]
    asset
  end

  def self.find_name_or_uuid_by_host(host)
    (host.respond_to?(:subscription_facet) && !host.subscription_facet.nil?) ? host.subscription_facet.try(:uuid) : host.name
  end

  private

  def self.find_host_by_name_or_uuid(cname)
    if defined?(Katello::Host::SubscriptionFacet)
      facet = Katello::Host::SubscriptionFacet.find_by_uuid(cname)
      host = facet.host if facet

      host ||= Host.find_by_name(cname)
    else
      host = Host.find_by_name(cname)
    end

    unless host
      Rails.logger.error "Could not find Host with name: #{cname}"
      fail ActiveRecord::RecordNotFound
    end
    host
  end
end
```

shlomizadok:
@ehelms - yeah, this is much better (and defensive). Changed

lib/foreman_openscap/helper.rb
```diff
@@ -16,25 +16,21 @@ def self.get_asset(cname, policy_id)
...
+    (host.respond_to?(:subscription_facet) && !host.subscription_facet.nil?) ? host.subscription_facet.try(:uuid) : host.name
   end
 
   private
 
   def self.find_host_by_name_or_uuid(cname)
-    if defined?(Katello::System)
-      host = Host.includes(:content_host).where(:katello_systems => {:uuid => cname}).first
+    if Facets.registered_facets.keys.include?(:subscription_facet)
+      host = Katello::Host::SubscriptionFacet.find_by_uuid(cname).host
```
ehelms: I think this still needs:

```
      facet = Katello::Host::SubscriptionFacet.find_by_uuid(cname)
      host = facet.host if facet
```

PR #182, opened 2016-07-21
Fixes #15756: Find by host object now that Katello systems are unified

ehelms:


lib/foreman_openscap/helper.rb
```diff
@@ -16,25 +16,15 @@ def self.get_asset(cname, policy_id)
...
   private
 
   def self.find_host_by_name_or_uuid(cname)
-    if defined?(Katello::System)
-      host = Host.includes(:content_host).where(:katello_systems => {:uuid => cname}).first
-      host ||= Host.find_by_name(cname)
-    else
-      host = Host.find_by_name(cname)
-    end
+    host = Host.find_by_name(cname)
```
shlomizadok: What happens when report is sending uuid as cname?

lib/foreman_openscap/helper.rb
```diff
@@ -16,25 +16,15 @@ def self.get_asset(cname, policy_id)
   end
 
   def self.find_name_or_uuid_by_host(host)
-    if defined?(::Katello::System)
-      host.content_host.uuid
-    else
-      host.name
-    end
+    host.name
```
ohadlevy: what about existing reports that use the cert uuid as their key? i think the test should check if there is a content facet and then use uuid or name?
also, its critical to find if this is used in migrations, as the content facet migration might happen after making this condition not relevant in migration.

shlomizadok:
This fixes migrated reports but breaks creating new ARF reports. Closing this in favor of #181 
Please re-open if you feel this needs further discussion. Thanks a lot @ehelms

PR #189, opened 2016-08-21
Fixes #16210 - policy update fail when associated hostgroup is deleted.

ranjan:


shlomizadok:
@ranjan - thanks for adding that!
I have two requests:
1. Will you add a test that covers this scenario, so we will not fail here next time
2. Fix commit message (see Foreman bot message above)
Thanks  :+1:

ranjan:
Thanks for quick reply @shlomizadok 
1. Fixed commit message.
2. I am not able to setup test env at my local :(. 
   But below unit test should work.
```
test "should remove associated hostgroup" do
    ForemanOpenscap::Policy.any_instance.stubs(:find_scap_puppetclass).returns(FactoryGirl.create(:puppetclass, :name => 'foreman_scap_client'))
    ForemanOpenscap::Policy.any_instance.stubs(:populate_overrides)
    hg1 = FactoryGirl.create(:hostgroup)
    asset = FactoryGirl.create(:asset, :assetable_id => hg1.id, :assetable_type => 'Hostgroup')
    policy = FactoryGirl.create(:policy, :assets => [asset])
    policy.hostgroup_ids = [hg1].map(&:id)
    policy.save!
    hg1.destroy
    assert_equal 0, policy.hostgroups.count   
  end
```

ares:
@ranjan if you just add it to a commit, it will be triggered by jenkins so you'll see whether it works.

ranjan:
@ares Test Added

shlomizadok:
Tested.  :+1: 
Will merge if no other comments.

app/models/concerns/foreman_openscap/hostgroup_extensions.rb
```diff
@@ -6,11 +6,16 @@ module HostgroupExtensions
       has_one :asset, :as => :assetable, :class_name => "::ForemanOpenscap::Asset"
       has_many :asset_policies, :through => :asset, :class_name => "::ForemanOpenscap::AssetPolicy"
       has_many :policies, :through => :asset_policies, :class_name => "::ForemanOpenscap::Policy"
+      after_destroy :delete_associated_hostgroup
```
ares: why not adding `dependent: destroy` to `has_one :asset` association instead?

shlomizadok: True.. missed that. A much better way to handle this

ranjan:
Thanks @ares,  PR updated.

PR #191, opened 2016-08-23
Fixes #16244 - change 'foreman' to 'tfm'

shlomizadok:
In javascript, change foreman to tfm


xprazak2:
ACK, merging. Thanks @shlomizadok!

PR #192, opened 2016-08-26
Fixes #16323 - restore widget rendering on policy page

ares:
I don't think it makes sense to load data by ajax on this page, therefore I just extracted original method implementation.


shlomizadok:
Makes senses. works as expected. Merging

PR #193, opened 2016-08-26
Fixes #16325 - policy delete should remove association

ranjan:


ares:
why is this desired? does it mean that deleting policy would delete hostgroup/host?

ranjan:
Added Fix for broken test of PR#189

ares:
I see now, I missed `:through` on which only the association is deleted. Since there can be left overs in this table already, could you add a migration that would delete those?

Also, is the test fix required because of this change? It seems broken for longer time right? In such case, could you please create a separate PR for that?

ares:
btw I tested the change, works as expected

ares:
Thanks, works nicely, merging! (tests failed only on https://github.com/theforeman/foreman_openscap/pull/197)

PR #196, opened 2016-08-29
Fixes #16361 - fix tests

ares:


ares:
Tested by @ranjan, merging so we get tests working again.

PR #201, opened 2016-08-31
Fixes #16387 - Fix apipie doc in policies controller

xprazak2:


shlomizadok:
APJ. Thanks @xprazak2 !

PR #203, opened 2016-09-02
Fixes #16431: Explicitly set path of PolicyApi include

ehelms:
Defining modules with syntax `::` has unintended consequences with
respect to how modules are looked up. This explicitly declares the
full path to the PolicyApi module instead of relying on Ruby to
attempt to automatically determine it.


shlomizadok:
:+1: 
APJ

ehelms:
This https://github.com/theforeman/foreman_openscap/pull/204 should fix the issue you were seeing with Foreman starting. The two changes were unrelated to each other.

xprazak2:
@shlomizadok it does not seem to break anything for me, does rebasing of core's develop help?

PR #204, opened 2016-09-03
Fixes #16432: set inherited_attributes to proper function name

ehelms:


xprazak2:
@ehelms I believe the method name is ok the way it is, it is meant to target [this](https://github.com/theforeman/foreman/blob/develop/app/models/host/managed.rb#L622) method. Purpose of `inherited_attributes` is to set the new values on ajax call when a hostgroup is changed:

https://github.com/theforeman/foreman/blob/develop/app/controllers/hosts_controller.rb#L653
https://github.com/theforeman/foreman/blob/develop/app/models/host/managed.rb#L617

You can see this when you create 2 hostgroups, one with `openscap_proxy` and the other without it. If you now try to create a new host, you will notice that Openscap Proxy dropdown in host form shows the value from the currently selected hostgroup.

But I agree it should be in `hostgroup_inherited_attributes` as well, could you make a change?

ares:
seems #206 addresses it?

ehelms:
Dunno how I missed that, I looked everywhere (I thought) for that defined method. My checkout must not have updated like I thought - closing as I think @ares has this covered.

PR #205, opened 2016-09-03
Fixes #16223: deleted compliance arf reports count

ranjan:


xprazak2:
Thanks for the fix, works great. Merging...

PR #206, opened 2016-09-05
Fixes #16443 - Add strong param filter for hostgroup

xprazak2:


ares:
Please create a separate issue for this so we can track the version, this can't get into 0.6.0 which is already released but will go to 0.6.1. Otherwise :+1:, fixes the issue for me.

ares:
Thanks @xprazak2, merging!

PR #208, opened 2016-09-07
Fixes #16366 - Add SCAP content validations to policy

shlomizadok:


test/unit/policy_test.rb
```diff
@@ -122,4 +122,22 @@ class PolicyTest < ActiveSupport::TestCase
...
+  test "should not create policy without SCAP content" do
+    p = ForemanOpenscap::Policy.new(:name => "custom_policy",
+                                    :scap_content_profile_id => @scap_profile.id,
+                                    :period => 'monthly',
+                                    :day_of_month => '0')
+    refute p.save
+    assert p.errors[:scap_content_id].include?("can't be blank")
+  end
+
+  test "should not create policy with monthly period" do
```
shlomizadok: Typo while copying... Will change once tests are finished

xprazak2:
Looks good, but there is no error message when I click 'next' without changing the dropdowns. Could you have another look please?

shlomizadok:
@xprazak2 - fixed

xprazak2:
Nice. Thanks!

xprazak2:
Note: I see JSON style hashes, so if all tests pass, then our rubocop is not running.

shlomizadok:
Oh, I thought it was changed (noticed it has changed in core)

xprazak2:
I knew there was a rubocop discussion, but I did not notice they actually changed it in core. It is still on in our [rubocop.yaml](https://github.com/theforeman/foreman_openscap/blob/master/.rubocop.yml#L15). There was also outdated rule when I tried running `rake foreman_openscap:rubocop` manually.

Seems like we can merge this and fix rubocop separately if there are no objections.

shlomizadok:
@xprazak2 - I believe this is ready to merge.
We should : A) Add houndci here and  rubocopify it :) & B) Allow ruby19 Hash syntax like in [core](https://github.com/theforeman/foreman/blob/develop/.rubocop.yml#L25-L29)

xprazak2:
Sounds like a plan. Merging, thanks!

PR #216, opened 2016-09-13
Fixes #16485 - don't rely on attribute being migrated

ares:


shlomizadok:
Merging. Thanks a lot, @ares !

PR #218, opened 2016-09-26
Fixes #16694 - Add endpoint for Arf report downloads

xprazak2:


config/routes.rb
```diff
@@ -55,7 +55,11 @@
             get 'content'
           end
         end
-        resources :arf_reports, :only => [:index, :show, :destroy]
+        resources :arf_reports, :only => [:index, :show, :destroy] do
+          member do
+            get 'download'
```
ares: I suggest we disable this cop

config/routes.rb
```diff
@@ -55,7 +55,11 @@
             get 'content'
           end
         end
-        resources :arf_reports, :only => [:index, :show, :destroy]
+        resources :arf_reports, :only => [:index, :show, :destroy] do
```
ares: I suggest we disable this cop, both syntax are fine

app/controllers/api/v2/compliance/arf_reports_controller.rb
```diff
@@ -62,12 +62,31 @@ def create
           render :json => { :result => :OK, :id => arf_report.id.to_s }
         end
 
+        api :GET, "/compliance/arf_reports/:id/download/", N_("Download bzipped ARF report")
```
ares: I suggest we disable this cop

config/routes.rb
xprazak2: I believe we have it [disabled](https://github.com/theforeman/foreman_openscap/blob/master/.rubocop.yml#L15-L17), is it possible hound is ignoring it?

config/routes.rb
xprazak2: Fine with me.

xprazak2:
I rebased and disabled the StringLiterals rule, hound seems to be satisfied.

ares:
One more thing is missing, you need to list the action for at least one permission, probably here - https://github.com/theforeman/foreman_openscap/blob/master/lib/foreman_openscap/engine.rb#L53

xprazak2:
I always seem to forget permissions...

shlomizadok:
We are missing test coverage for this. Will you please add?

xprazak2:
I rebased and added a test case.

shlomizadok:
Merging. Thanks a lot, @xprazak2 !

PR #224, opened 2016-11-09
Fixes #16878 - Add button leading to full report

xprazak2:


ares:
Thanks @xprazak2, merging!

PR #227, opened 2016-12-13
Fixes #17239 - Increase scap content validation timeout

shlomizadok:


app/lib/proxy_api/openscap.rb
```diff
@@ -46,5 +49,11 @@ def destroy_report(report, cname)
         false
       end
     end
+
+    private
+
+    def timeout
+      Setting[:proxy_request_timeout] > 120 ? Setting[:proxy_request_timeout] : 120
```
xprazak2: I think you need something like `Setting[:proxy_request_timeout] && Setting[:proxy_request_timeout] > 120` for tests to pass.

shlomizadok:
test failures are unrelated

xprazak2:
Seems to take care of timeout issue for me, but my proxy gets killed by kernel 2 out of 3 times because I seem to be out of memory. This may be because I am running a devel proxy in a box, but I wonder if we should handle this somehow. The exceptions are `RestClient::ServerBrokeConnection` when killed during validation, `Errno::ECONNREFUSED` when `fetch_policies_for_scap_content` is called. Your thoughts?

shlomizadok:
Weird... it didn't happen to me with the file from the issue. I also run devel proxy (on my machine - not in a box...)

xprazak2:
I just re-tested in different env and it worked flawlessly. The issues I had above seem to be tied to my setup. Jenkins failures unrelated, seems like ready to merge if there are no further comments.

xprazak2:
Merged, thanks @shlomizadok!

PR #229, opened 2016-12-22
Fixes #17831 - Add resource_type to ARF permission mapping

shlomizadok:


ares:
looks good to me, I suppose this started to be an issue when STI was introduced to reports?

shlomizadok:
@ares, yeah... especially around name space change from :scaptimony_arf_reports

ares:
It does not seem to work for me, my permissions do not have resource_type specified and restarting or seeding does not modify them

```
irb(main):008:0> Permission.where :name => 'view_arf_reports'
2016-12-23T09:22:41  [sql] [D]   Permission Load (0.4ms)  SELECT "permissions".* FROM "permissions" WHERE "permissions"."name" = $1  [["name", "view_arf_reports"]]
+-----+------------------+---------------+-------------------------+-------------------------+
| id  | name             | resource_type | created_at              | updated_at              |
+-----+------------------+---------------+-------------------------+-------------------------+
| 204 | view_arf_reports |               | 2016-04-28 06:49:41 UTC | 2016-04-28 06:49:41 UTC |
+-----+------------------+---------------+-------------------------+-------------------------+
```

do we need a migration for existing systems? Note that the permission with changed resource type won't be created because the old with the same name already exists https://github.com/theforeman/foreman/blob/develop/app/services/foreman/plugin.rb#L232

shlomizadok:
@ares, indeed we need a migration :thinking: - added.

ares:
I don't think the migration works, it does nothing if it finds existing record with such name. You should probably use Permission.update_all anyway to avoid future validations conflicts.

db/migrate/20161223153249_add_permissionsto_arf_report.rb
```diff
@@ -0,0 +1,13 @@
+class AddPermissionstoArfReport < ActiveRecord::Migration
+  def up
+    %(view_arf_reports destroy_arf_reports).each do |name|
```
ares: %w I guess...

shlomizadok:
@ares - fixed

ares:
hound

ares:
The migration uses `update_attributes!` which is an alias for `update!`. These methods trigger validations. The migration must skip validations, please use class method `update_all`. While this is not a problem now, later when new validation is added on column that would be added in migration after this migration, this migration would fail.

shlomizadok:
@ares, thanks for the comment. Changed to `update`

ares:
@shlomizadok this is still not correct, please use [class method update_all](http://apidock.com/rails/ActiveRecord/Relation/update_all) to avoid triggering any validations on the existing model instance, see example [here](https://github.com/theforeman/foreman/blob/develop/db/migrate/20161007115719_rename_ipmi_boot_permission.rb#L3). If we need to create new models we must use fake model, example can be found [here](https://github.com/theforeman/foreman/blob/698e916ce208b5040b83a908a058c83c94d158ee/db/migrate/20140908192300_change_nil_admin_users_to_false.rb#L1). Just define this fake model inside the migration class.

Otherwise the migration is not safe for future.

db/migrate/20161223153249_add_permissionsto_arf_report.rb
```diff
@@ -0,0 +1,11 @@
+class AddPermissionstoArfReport < ActiveRecord::Migration
```
ares: s/AddPermissionstoArfReport/AddPermissionsToArfReport/

also the filename needs underscore

db/migrate/20161223153249_add_permissions_to_arf_report.rb
```diff
@@ -0,0 +1,11 @@
+class AddPermissionsToArfReport < ActiveRecord::Migration
```
shlomizadok: woof

ares:
@shlomizadok could you please rebase this? This check was already disabled but your branch does not contain it.

shlomizadok:
@ares :+1:

ares:
APJ

ares:
[test]

ares:
Thanks @shlomizadok, merging.

PR #232, opened 2017-01-04
Fixes #17464 - Use digest to detect scap_file changes

xprazak2:


shlomizadok:
@xprazak2 - rebase please :)

xprazak2:
Rebased

xprazak2:
[test]

ares:
Works, before we merge, I wonder if we should do one smart_proxy_openscap release before we have this in. I suppose users will first upgrade their Foreman instances and of the Foreman configures new download path, smart proxy won't be able to serve it until updated, correct?

xprazak2:
Correct, if foreman_openscap is updated and configures a client with a new `download_path` (containing a hash), then proxy will not be able to serve unless updated as well.

ares:
Thanks @xprazak2, merging

PR #235, opened 2017-01-17
Fixes #17240 - Add spool failure warning

xprazak2:


lib/foreman_openscap/engine.rb
```diff
@@ -104,13 +104,15 @@ class Engine < ::Rails::Engine
           permission :destroy_tailoring_files, { :tailoring_files => [:destroy],
                                                  'api/v2/compliance/tailoring_files' => [:destroy] },
                       :resource_type => 'ForemanOpenscap::TailoringFile'
+          permission :openscap_spool, { :smart_proxies => [:openscap_spool] }
```
ares: what is this permission used for if we map the action to view smart proxy permission? https://github.com/theforeman/foreman_openscap/pull/235/files#diff-8998f9b7f683bfc0bcb3fe387b588f44R23

if we want to keep it, shouldn't it have a resource specified?

ares:
One nit otherwise looks good.

xprazak2:
I removed the new permission, `:view` on smart proxy will work

ares:
[test]

ares:
It seems that removing of that permission broke tests

```
AccessPermissionsTest.test_0881_route smart_proxies/openscap_spool should have a permission that grants access

/var/lib/workspace/workspace/test_plugin_pull_request@2/database/mysql/label/fast/ruby/2.0.0/foreman/test/unit/shared/access_permissions_test_base.rb:34
```

could you please take a look?

xprazak2:
I moved things to a separate controller, hopefully it will pass permissions test this time.

app/views/smart_proxies/_openscap_spool.html.erb
```diff
@@ -0,0 +1,7 @@
+<span>
+  <% if last_error %>
+    <%= _('%s ago') % time_ago_in_words(Time.at(last_error["timestamp"])) %>
```
ares: the time information does not say much, could you please add a link to logs? this seems to work just fine
```
<%= link_to _('See logs for more information', '#logs') %>
```

also with following js you could filter relevant logs

```js
var string = 'Failed to parse Arf Report';
$("div#table-proxy-status-logs_filter input").val(string);
var table = $('#table-proxy-status-logs').DataTable();
table.search(string).draw();
```

It would be also good to add a help icon with tooltip explaining what is "spool error". Bonus points for sending a PR to core that would add a JS function so plugins could generate such links easier.

xprazak2:
I added the tooltip with a hint and a link that leads to logs with spool failure.

app/views/smart_proxies/plugins/_openscap.html.erb
```diff
@@ -0,0 +1,12 @@
+<%= javascript 'foreman_openscap/openscap_proxy' %>
+<div class="row">
+  <h3><%= feature %></h3>
+</div>
+<%= show_feature_version(feature.downcase) %>
+<div class="row">
+  <% msg = _("Proxy failed to send a report from spool to Foreman. This indicates a corrupted report format. Report has been moved to :corrupted_dir on proxy for inspection.") %>
```
ares: I'd change `Report has been moved to :corrupted_dir on proxy for inspection` to `Report has been moved to directory for storing corrupted files on proxy for later inspection.`

xprazak2:
I changed the message as suggested.

ares:
ACK pending jenkins

ares:
Thanks @xprazak2 merging.

PR #237, opened 2017-01-24
Fixes #18207 - Disable tailoring feature for old proxies

xprazak2:
I tried to make this as simple as possible. Behavior:

Tailoring files are disabled in UI (you can't create one) if there is a proxy with 'Openscap' feature and version is lower than required. If any of the openscap proxies is down, user is asked to make them available so that foreman can check the version. After the check passes, the feature is enabled.

~~API and CLI only checks if proxies have required versions when creating a tailoring file and ignores those that are down. Require all openscap proxies to be up when creating a tailoring file seems too harsh. I plan to add a warning banner to hammer.~~

API will check when there are no tailoring files as per today's lunch discussion.


ares:
[test]

ares:
[test]

ares:
Thanks @xprazak2, merging!

PR #238, opened 2017-01-24
Fixes #18218 - Use proxy version to determine download path

xprazak2:
Will need #237 and #232 to be applied first.

### DIFF
--- a/app/models/concerns/foreman_openscap/host_extensions.rb
+++ b/app/models/concerns/foreman_openscap/host_extensions.rb
@@ -63,7 +63,9 @@ module ForemanOpenscap
     end
 
     def policies_enc
-      combined_policies.map(&:to_enc).to_json
+      check = ForemanOpenscap::OpenscapProxyAssignedVersionCheck.new(self).run
+      method = check.pass? ? :to_enc : :to_enc_legacy
+      combined_policies.map(&method).to_json
     end
 
     def combined_policies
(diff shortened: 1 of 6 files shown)


app/models/concerns/foreman_openscap/host_extensions.rb
```diff
@@ -63,7 +63,9 @@ def get_asset
     end
 
     def policies_enc
-      combined_policies.map(&:to_enc).to_json
+      check = ForemanOpenscap::OpenscapProxyAssignedVersionCheck.new(self).run
```
ares: The check in here checks for `<= 0.6.1` but 0.6.1 already has tailoring file API and digests. The comparing should be `< 0.6.1` I believe

ares:
```diff
diff --git a/app/services/foreman_openscap/openscap_proxy_version_check.rb b/app/services/foreman_openscap/openscap_proxy_version_check.rb
index 0be0c37..8c67098 100644
--- a/app/services/foreman_openscap/openscap_proxy_version_check.rb
+++ b/app/services/foreman_openscap/openscap_proxy_version_check.rb
@@ -9,7 +9,7 @@ module ForemanOpenscap
 
     def run
       @versions = openscap_proxy_versions.select do |key, value|
-        Gem::Version.new(value) <= Gem::Version.new("0.6.1")
+        Gem::Version.new(value) < Gem::Version.new("0.6.1")
       end
       self
     end
```

this should do the trick

xprazak2:
Updated with fixes, hopefully we'll get green form Jenkins

ares:
Thanks @xprazak2, merging.

PR #240, opened 2017-02-14
Fixes #18501 - Align with core to fix tests

xprazak2:


ares:
Thanks @xprazak2, merging!

PR #241, opened 2017-02-23
Fixes #18100 - Download Arf report as HTML

xprazak2:
This is the smallest change to implement report downloads in HTML

pros:
* very small, easy cherrypick for downstream
* users can download report formats separately
* no changes needed on proxy

cons:
* users have to click twice if they want XML and HTML
* additional button in UI

If we wanted to add HTML into existing bzip archive with XML, we would have to work with bzip on proxy, which means either use `system` call (ugly, how do we guarantee bzip is really there?) or add something like [bzip2-ruby](https://github.com/brianmario/bzip2-ruby) into dependencies (packaging change). I would go with this solution for now and raise refactoring issue if we really want to download both formats in one archive.


xprazak2:
I updated the bzip files with new names.

xprazak2:
I added the file extensions for XML files.

ares:
Works nicely, big :+1: since it also improves the xml download usability. Can be merged once mr. Jenkins confirms.

shlomizadok:
@xprazak2 - amazing !! :+1:  :+1: :tada: 
@ares - may I merge this one? :smile:

shlomizadok:
Merging (Mr. Jenkins has gone fishing)

shlomizadok:
@xprazak2 !!!!!!!!!
(Thanks !)

ares:
Thanks for merging, please don't forget to set the right release in redmine after you merge something.

PR #242, opened 2017-03-01
Fixes #18734 - Use Rex feature to run scans

xprazak2:


xprazak2:
This may also be useful to do in bulk from hosts/index page. Is remote execution planning on supporting it in some way or should I add a button?

ares:
There's already a bulk action "Run Job" that takes you to REX form. There's just one manual step to select the job you want to run, in this case it would be scap scan. Personally I think that is good enough, I'd like to avoid adding a lot more custom bulk actions to the menu that already contains more than 10 items.

xprazak2:
OK, sounds reasonable.

ares:
Somehow tests seem to be broken by this, but form quick scan code looks good.

ares:
[test]

ares:
Tests are green, thanks @xprazak2, merging!

PR #243, opened 2017-03-03
Refs #18734 - improve feature name

ares:
I'm not providing a migration since there was no release with this feature yet. @xprazak2 could you review pls?


ares:
also the migration is not needed, since the REX updates the feature automatically :-)

xprazak2:
Merging, thanks!

PR #244, opened 2017-03-06
Fixes #18809 - simplify the search function

ares:


xprazak2:
Takes case of the issue for me, merging. Thanks!

PR #245, opened 2017-03-06
Fixes #18810 - use relation instead of in

ares:


app/models/concerns/foreman_openscap/host_extensions.rb
```diff
@@ -10,13 +10,13 @@ module HostExtensions
...
       has_one :compliance_status_object, :class_name => '::ForemanOpenscap::ComplianceStatus', :foreign_key => 'host_id'
 
-      scoped_search :in => :policies, :on => :name, :complete_value => true, :rename => :compliance_policy,
-                    :only_explicit => true, :operators => ['= ', '!= '], :ext_method => :search_by_policy_name
+      scoped_search :relation => :policies, :on => :name, :complete_value => true, :rename => :compliance_policy,
+                    :only_explicit => true, :operators => ['= '], :ext_method => :search_by_policy_name
 
-      scoped_search :in => :policies, :on => :name, :complete_value => true, :rename => :compliance_report_missing_for,
+      scoped_search :relation => :policies, :on => :name, :complete_value => true, :rename => :compliance_report_missing_for,
                     :only_explicit => true, :operators => ['= ', '!= '], :ext_method => :search_by_missing_arf
```
xprazak2: Should `!=` be removed here as well because it was removed on line 14?

ares: I don't know, I haven't tested whether it works correctly, I'm just sure the first one does not

xprazak2: Tested, it works correctly for me, so I think we can leave it as it is.

xprazak2:
Merging, thanks!

PR #246, opened 2017-03-08
Fixes #18837 - Delete reports when host is missing

xprazak2:
I could not reproduce this issue because all reports get [deleted with host](https://github.com/theforeman/foreman/blob/develop/app/models/host/managed.rb#L21). I am not sure how this problem can even happen because reports have foreign key constraint on host_id in db. Anyway, this should delete the reports and and avoid a crash.


ares:
Thanks @xprazak2, merging.

PR #247, opened 2017-03-08
Fixes #18841 - Added policies to hosts index api scope

ShimShtein:
Policies are loaded while calculating compliance status. To avoid N+1 query, we are preloading them.


xprazak2:
@ShimShtein, it works nicely. Could you add the same for hosts UI controller as well?

ShimShtein:
We don't need it in the UI controller. At least I couldn't see the query there.

xprazak2:
That is strange, I can see the query:

ShimShtein:
Added definition for UI controller too.

xprazak2:
Working nicely, thanks!

PR #248, opened 2017-03-23
Fixes #19007 - limit supported REX version

ares:


ares:
@xprazak2 could you please review? we need to release 0.6.6 for 1.14 soon, all users that use foreman_openscap and foreman_remote_execution with 1.14 can't start Foreman

lib/foreman_openscap/engine.rb
```diff
@@ -155,7 +155,7 @@ class Engine < ::Rails::Engine
         parameter_filter Hostgroup, :openscap_proxy_id, :openscap_proxy
         parameter_filter Log, :result
 
-        if ForemanOpenscap.with_remote_execution?
+        if ForemanOpenscap.with_remote_execution? && Gem::Version.new(ForemanRemoteExecution::VERSION) >= Gem::Version.new('1.2.3')
```
stbenjam: Is there any reason to have the feature if the version is less than 1.2.3? Because 1.2.2 and earlier can register features just not host buttons.

ares: It's not used anywhere else AFAIK.

stbenjam: Ah ok, LGTM then

xprazak2:
Merging, thanks!

PR #251, opened 2017-04-04
fixes #19164 - create feature even if on older rex

stbenjam:
The template references the feature, so it always needs to exist.


ares:
note for myself: we'll need to cherry pick this to 0.6 branch for 1.14

ares:
Thanks @stbenjam, tested with older version of REX and it works just fine. After updating, the host_action_button is correctly set to true after re-seeding the db. Merging and will do the releases for both 1.14 and 1.15.

stbenjam:
Thanks much! I also found another problem when testing on older Foreman, the smart proxy has a 1.8.7 incompatibility, there's a PR over there https://github.com/theforeman/smart_proxy_openscap/pull/48

PR #252, opened 2017-04-11
Fixes #19242 - Unscope smart proxies

xprazak2:


ares:
I think a slightly better solution would be to set the `User.current = User.anonymous_admin` in the rake task, like we did in https://github.com/theforeman/foreman_templates/pull/46/files

Otherwise users might be using proxies from different organization when they import new content files from UI if I understood it correctly..

xprazak2:
`User.current = User.anonymous_admin` takes care of the issue.

ares:
[test]

ares:
Thanks @xprazak2, merging!

PR #254, opened 2017-05-04
Fixes #19462 - changed includes to preload

ShimShtein:


shlomizadok:
LGTM, pending Jenkins

ares:
Thanks @ShimShtein, do you have some testing data or something where I could reproduce the issue? My understanding is that `preload` enforces always uses 2 SQL queries, while `includes` can fallback to `LEFT OUTER JOIN` in case there's some condition on the joined table. This change would make the `base_scope` limited in term of adding `where` with conditions on policies table. Also I was under impression, that `include` does the same things as `preload` (2 queries) as long as there's no such condition , so I wonder how this helps with memory consumption.

iNecas:
FYI https://bugzilla.redhat.com/show_bug.cgi?id=1447958#c13 confirms this change really has impact on the memory. The issue seems to be that the `include` generates too big queries to the rdbms to handle

ShimShtein:
@iNecas  already stated the reason. If you want a bit more info, I have explained this issue in Katello's part of this chage [here](https://github.com/Katello/katello/pull/6774#issuecomment-299197260).

Basically, I didn't see a problem with policies, but I am not sure it couldn't create problems in some esoteric use case somewhere down the road. IMHO using preload (two queries) is justified, if there is a big degree of connectivity (many hosts are connected to the same small set of policies).

ares:
Thanks @ShimShtein, works fine with hosts that have policies. I'm merging since it's unlikely that this would break testing. I'll fix them if that is the case :-)

PR #258, opened 2017-05-16
Fixes #19552 - Fix search for hosts without arf report

xprazak2:
I think these queries would deserve some tests. The problem is, they need joins across multiple models which would mean setting up associations in our factories and make changes in majority of our tests.

### DIFF
--- a/app/models/concerns/foreman_openscap/host_extensions.rb
+++ b/app/models/concerns/foreman_openscap/host_extensions.rb
@@ -35,15 +35,11 @@ module ForemanOpenscap
       }
 
       scope :policy_reports_missing, lambda { |policy|
-        where("id NOT IN (SELECT host_id
-                          FROM reports INNER JOIN foreman_openscap_policy_arf_reports
-                              ON reports.id = foreman_openscap_policy_arf_reports.arf_report_id
-                          WHERE policy_id = #{policy.id})
-              AND id IN (SELECT assetable_id
-                         FROM foreman_openscap_asset_policies INNER JOIN foreman_openscap_assets
-                              ON foreman_openscap_asset_policies.asset_id = foreman_openscap_assets.id
-                         WHERE foreman_openscap_assets.assetable_type = 'Host::Base'
-                               AND foreman_openscap_asset_policies.policy_id = '#{policy.id}')")
+        search_for("compliance_report_missing_for = \"#{policy.name}\"")
+      }
+
+      scope :assigned_to_policy, lambda { |policy|
+        search_for("compliance_policy = \"#{policy.name}\"")
       }
 
       alias_method_chain :inherited_attributes, :openscap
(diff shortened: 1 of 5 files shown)


ares:
[test]

xprazak2:
[test]

xprazak2:
[test]

ares:
Failures unrelated, the branch does not contain e1d06d9db38c33d4454acf1d4e72eb3670f78d11

app/models/concerns/foreman_openscap/host_extensions.rb
```diff
@@ -109,15 +101,13 @@ def search_by_policy_name(key, operator, policy_name)
...
-        { :conditions => Host::Managed.arel_table[:id].in(Host::Managed.select(Host::Managed.arel_table[:id])
-            .joins(:policies)
-            .where(cond)
-            .where("foreman_openscap_assets.id NOT IN (
-                     SELECT DISTINCT foreman_openscap_arf_reports.asset_id
-                     FROM foreman_openscap_arf_reports
-                     WHERE foreman_openscap_arf_reports.asset_id = foreman_openscap_assets.id
-                         AND foreman_openscap_arf_reports.policy_id = foreman_openscap_policies.id)")
-            .pluck(:id)).to_sql}
+        host_ids_from_arf_of_policy = ForemanOpenscap::ArfReport.joins(:policy).where(cond).pluck(:host_id).uniq
```
ares: This seems to only search for direct Host - Policy associations, if the host has policy because it's assigned to host group, it's not in the list. It seems it didn't work previously either, could it be incorporated in the patch?

ares:
It seems to work well, just one comment that would be nice to fix too.

app/models/concerns/foreman_openscap/host_extensions.rb
xprazak2: It was working when I tested, the policy is never connected directly to the host - it is linked through Asset or ArfReport. It does not matter if a certain policy comes from a hostgroup - it is still linked to a host through associated asset.

xprazak2:
It should be working for policies inherited from hostgroups now.

xprazak2:
[test]

ares:
Counts should also change at https://github.com/theforeman/foreman_openscap/blob/master/app/services/foreman_openscap/policy_dashboard/data.rb#L25-L26, we're ignoring host associated through host group there. It should probably use the same scope `search_by_policy_name` to which it links to. That also needs to be fixed in a same way https://github.com/theforeman/foreman_openscap/blob/master/app/models/concerns/foreman_openscap/host_extensions.rb#L107 :-( I've sent you few suggestions on how to fix it.

xprazak2:
I made changes as suggested.

ares:
Works nicely, thanks @xprazak2, merging!

PR #259, opened 2017-05-16
Fixes #19562 - Fix failing tests after minitest update

xprazak2:


ares:
Job triggered at http://ci.theforeman.org/job/test_plugin_pull_request/3691/, it seems mostly green but on one combination it has "ActiveRecord::RecordInvalid: Validation failed: Name has already been taken". I've trigger the second run at http://ci.theforeman.org/job/test_plugin_pull_request/3692/ just to confirm, it's intermittent.

ares:
Now it seems the same error occured on 2 different combinations so we might need to fix that too, but for now I think we can merge these fixes. Thanks @xprazak2, merging.

PR #260, opened 2017-05-17
Refs #19562 - Avoid clash with core's factories

xprazak2:
I managed to get the same error locally when running tests multiple times. The errors we saw on Jenkins were all caused by openscap_proxy factory, so maybe it somehow clashes with the factory from core. I changed the proxy test data to be different and run tests locally 30 times without an error.


ares:
Thanks @xprazak2, merging!

PR #261, opened 2017-05-22
Fixes #19621 - restore compatibility with new REX

ares:


ares:
errors seems to be unrelated

db/seeds.d/75-job_templates.rb
```diff
@@ -3,7 +3,12 @@
     JobTemplate.without_auditing do
       Dir[File.join("#{ForemanOpenscap::Engine.root}/app/views/job_templates/**/*.erb")].each do |template|
         sync = !Rails.env.test? && Setting[:remote_execution_sync_templates]
-        JobTemplate.import!(File.read(template), :default => true, :locked => true, :update => sync)
+        # import! was renamed to import_raw! around 1.3.1
+        if JobTemplate.respond_to?('import_raw!')
+          JobTemplate.import_raw!(File.read(template), :default => true, :locked => true, :update => sync)
+        else
+          JobTemplate.import(File.read(template), :default => true, :locked => true, :update => sync)
```
xprazak2: as per comment on line 6, `import!` was renamed to `import_raw!` but there is only `import` here (without a bang)

ares: good catch, I wonder which fat finger deleted the bang :-) fixed

xprazak2:
Merged, thanks!

PR #263, opened 2017-06-01
Fixes #19629 - Fix openscap proxy inheritance in UI

xprazak2:


app/models/concerns/foreman_openscap/hostgroup_extensions.rb
```diff
@@ -16,6 +16,15 @@ def inherited_policies
       end.uniq
     end
 
+    def openscap_proxy
+      return super if ancestry.nil? || self.openscap_proxy_id.present?
+      ::SmartProxy.find_by(:id => inherited_openscap_proxy_id)
+    end
+
+    def inherited_openscap_proxy_id
+      inherited_ancestry_attribute(:openscap_proxy_id)
```
ares: This method is for some reason only defined `unless defined?(Katello::System)`, if we want to use it here, we should make it always available. It seems that you added it some time ago, do you remember why the condition was required?

ares:
except the method scope looks good and works fine

app/models/concerns/foreman_openscap/hostgroup_extensions.rb
xprazak2: Because I 'borrowed' this from [Katello](https://github.com/Katello/katello/blob/master/app/models/katello/concerns/hostgroup_extensions.rb#L72). If there is no Katello, we define it ourselves, if we run with Katello, no need to define it again.

ares: Shouldn't this method live in core then? That does not block this PR but such small refactoring would be nice (add it to core and remove from katello and openscap plugins)

ares:
[test] any idea why test fail? it does not seem related

ares:
Thanks @xprazak2, merging! The followup refactoring of moving the method into core would be appreciated.

PR #264, opened 2017-06-06
Fixes #19705 - Do not look for spool errors when logs disabled

xprazak2:


ares:
[test]

app/views/smart_proxies/_openscap_spool.html.erb
```diff
@@ -1,8 +1,12 @@
 <span>
-  <% if last_error %>
+  <% if last_error && last_error.has_key?('timestamp') %>
     <span>
       <a href="#logs" onclick="findSpoolLogs();"><%= _('%s ago') % time_ago_in_words(Time.at(last_error["timestamp"])) %></a>
     </span>
+  <% elsif last_error.empty? %>
+    <span>
+      <%= _('Enable "logs" feature on proxy to search for errors') %>
```
ares: I'm seeing this even if last_error is nil (logs enabled but no error found), which is caused by the fact that `nil.empty?` returns `true`. It should probably also verify `!last_error.nil?`. This happens because of https://github.com/theforeman/foreman/blob/develop/lib/core_extensions.rb#L3 :-(

ares:
Thanks @xprazak2, merging. Test failures unrelated.

PR #268, opened 2017-06-26
Fixes #20096 - render charts again

ares:


ares:
@xprazak2 would you mind to look? should be small, I'd like to do another release with this fix soon.

shlomizadok:
APJ (I know that u missed me... 😉 )

ares:
[test]

xprazak2:
Looking good, but I think we need one additional change on [the reports page](https://github.com/theforeman/foreman_openscap/blob/master/app/views/arf_reports/show.html.erb#L1)

ares:
Good catch, added, please take another look.

xprazak2:
Merging, thanks!

PR #270, opened 2017-06-29
Fixes #18741 - Fix filtering of report messages

xprazak2:


ares:
actually it seems to be problem in core as well, we could fix it there and keep using core method

xprazak2:
@ares, I know we wanted to fix this in core, but core does not use multiple labels in the table, so it does not break core because of the way it is used and we also filter on different labels - warning, danger, success while core has info, default, warning, danger. But I can try to get this in core if you want me to.

ares:
I think that reusing the code from the core is a small victory, I understand it's not reproducible upstream but I think no one would object adding extra css class to this column and update the JS function to use it. The amount of duplication here seems too high.

app/helpers/arf_reports_helper.rb
```diff
@@ -22,7 +22,7 @@ def result_tag(level)
           else
             "warning"
           end
-    "class='label label-#{tag}'".html_safe
+    "class='label label-#{tag} result-filter-tag'".html_safe
```
ares: This is all we need from this PR now after https://github.com/theforeman/foreman/pull/4728 was merged. I already tested it works, please remove the rest.

xprazak2:
I kept only the newly added css class, removed the rest.

ares:
Thanks @xprazak2, merging!

PR #274, opened 2017-08-14
Fixes #20599 - show html guide fixes

ares:
found few more things that I'd like to separate so keeping as WIP

### DIFF
--- a/app/models/foreman_openscap/policy.rb
+++ b/app/models/foreman_openscap/policy.rb
@@ -1,3 +1,4 @@
+require 'rack/utils'
 module ForemanOpenscap
   class Policy < ApplicationRecord
     include Authorizable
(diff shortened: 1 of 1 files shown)


ares:
Requires https://github.com/theforeman/smart_proxy_openscap/pull/54 to work for default xccdf profiles.

xprazak2:
I was looking into the Default profile, it turns out to be poorly named [blank value](https://github.com/theforeman/foreman_openscap/blob/master/app/helpers/policies_helper.rb#L42). I do not know why we even have this and how it is supposed to work, it shows only when scap content has more than one profile. Besides, using it for scan results in empty report.

I think we can change it to 'Choose XCCDF profile' and remove the branching in `scap_content_profile_selector`.

ares:
Turns out blank value is default profile which is always present, just not frequently used. So it's a valid value.

app/models/foreman_openscap/policy.rb
```diff
@@ -28,7 +29,7 @@ class Policy < ApplicationRecord
               :if => Proc.new { |policy| policy.should_validate?('Schedule') }
 
     validates :scap_content_id, presence: true, if: Proc.new { |policy| policy.should_validate?('SCAP Content') }
-    validates :scap_content_profile_id, presence: true, if: Proc.new { |policy| policy.should_validate?('SCAP Content') }
+    #validates :scap_content_profile_id, presence: true, if: Proc.new { |policy| policy.should_validate?('SCAP Content') }
```
xprazak2: Maybe uncomment this and fix the default profiles separately? It should turn the tests green.

ares:
@xprazak2 updated the validation, please take another look. Also the proxy PR should be merged first.

xprazak2:
Merging, thanks!

PR #275, opened 2017-08-15
Fixes #20606 - Add missing rabl template for scap content

xprazak2:


ares:
Looks fine, could we print title instead of name? Now it only prints message and id. Or is that a hammer-cli-openscap change?

xprazak2:
Yes, some changes on hammer side are definitely needed to print title. For csv, we would need to override the [print_message](https://github.com/theforeman/hammer-cli/blob/master/lib/hammer_cli/output/adapter/csv.rb#L174-L196) from csv output adapter.

xprazak2:
Yes, some changes on hammer side are definitely needed to print title. For csv, we would need to override the [print_message](https://github.com/theforeman/hammer-cli/blob/master/lib/hammer_cli/output/adapter/csv.rb#L174-L196) from csv output adapter.

ares:
Ok, I wonder why `name` was not used in the first place... It would work if we override the method in CreateCommand like this
```ruby
def print_message(msg, msg_params={})
        super(msg, msg_params)
      end
```

But it would contain header saying `Name` so it would be confusing. Let's keep it like this for now, I opened a bug on hammer-cli http://projects.theforeman.org/issues/20744 and merging as it is.

Thanks @xprazak2!

PR #276, opened 2017-08-21
Fixes #20655 - rename notification policy

ares:


xprazak2:
Works for me, merging. Thanks!

PR #279, opened 2017-08-21
Fixes #20666 - Check for profile and content mismatch in policy

xprazak2:


ares:
Seems like tests needs to be updated, otherwise :+1: on the direction

ares:
Thanks @xprazak2, merging!

PR #282, opened 2017-08-21
Fixes #16490 - extend hostgroup and host API

ares:
@xprazak2 mind to take a look? Seems like we can use this http://projects.theforeman.org/projects/foreman/wiki/How_to_Create_a_Plugin#Adding-a-Smart-Proxy

with this we even get extended rabl templates

### DIFF
--- a/app/models/concerns/foreman_openscap/openscap_proxy_extensions.rb
+++ b/app/models/concerns/foreman_openscap/openscap_proxy_extensions.rb
@@ -2,10 +2,6 @@ module ForemanOpenscap
   module OpenscapProxyExtensions
     extend ActiveSupport::Concern
 
-    included do
-      belongs_to :openscap_proxy, :class_name => "SmartProxy"
-    end
-
     def openscap_proxy_api
       return @openscap_api if @openscap_api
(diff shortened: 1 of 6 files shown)


xprazak2:
It is as you suspected, [Katello](https://github.com/Katello/katello/blob/master/app/controllers/katello/concerns/api/v2/hostgroups_controller_extensions.rb#L7) overrides the doc. 

We should probably move our own [belongs_to](https://github.com/theforeman/foreman_openscap/blob/master/app/models/concerns/foreman_openscap/openscap_proxy_extensions.rb#L6) just to ArfReport.

ares:
Works with https://github.com/Katello/katello/pull/6937

xprazak2:
Works nicely, waiting for Katello PR to be merged.

xprazak2:
@ares, I think we should remove the `:belongs_to` from `OpenscapProxyExtensions` and move it just to `ArfReport`. `BelongsToProxies` sets up the relations to `Host` and `Hostgroup` for us when registering in `Engine`, see my comment above.

ares:
You're right, good catch. Updated.

xprazak2:
Merging, thanks!

PR #285, opened 2017-08-30
Fixes #20811 - add index at logs.result

iNecas:


ares:
Thanks @iNecas, merging.

PR #286, opened 2017-09-01
Fixes #20828 - Handle scap content profile title update

xprazak2:


ares:
Thanks @xprazak2, merging

PR #287, opened 2017-09-01
0.7.5 cherrypicks

ares:
@xprazak2 these all should be safe to cherry-pick for stable branch, could you please give it sanity check and then we can merge and release 0.7.5


xprazak2:
We should probably have #288 here as well.

ares:
@xprazak2 added two more commits, anything else on your mind?

xprazak2:
Nothing else on my mind, seems like we are good here.
Merging, thanks!

PR #288, opened 2017-09-04
Refs #20671 - Hostgroups in policy API

xprazak2:


ares:
Thanks @xprazak2, merging.

PR #290, opened 2017-09-14
Fixes #20930 - reenable default profiles

ares:
@xprazak2 it seems to work without changes on proxy side


ares:
Tests are broken on Ruby 2.3+, since that runs on Rails 5. Tests manually triggered at http://ci.theforeman.org/job/test_plugin_matrix/3729/

ares:
I'm taking the risk and merging without ACK. Reasoning:
1) @xprazak2 who's right now the only person doing reviews already saw the patch in other PR and tested it there, he will be offline for a week
2) I prefer to include it into today release so it does not have to wait for 2 weeks
3) tests with Rails 4 stack are green
4) it's trivial patch

PR #291, opened 2017-09-19
Fixes #20444 - Add validations for uniqueness content title

shiramax:


xprazak2:
[test]

ares:
Tests won't run until we get master tests stable

xprazak2:
@shiramax: I think you'll need to make a change in the [factory](https://github.com/theforeman/foreman_openscap/blob/master/test/factories/scap_content_related.rb#L5), something like:

```ruby
f.sequence(:title) { |n| "scap_content_title_#{n}" }
``` 

could work.

shiramax:
@xprazak2 Done, I also added migrations for the title uniqueness.

xprazak2:
[test]

ares:
Failures seems to be caused by new rails 5 migration syntax. `Migration[5.0]` does not seem to work, should the 5.0 tag be removed?

shiramax:
@ares Done.

xprazak2:
[test]

shiramax:
[test]

xprazak2:
The migration error is probably not intermittent, when I looked at my db, I got:



the table was renamed in the past but the index names were not, that seems to cause the migration to fail now. I would try to get the index name from table and then remove it by name.

xprazak2:
[test]

shiramax:
[test]

xprazak2:
[test]

ares:
Test failures unrelated, LGTM. Leaving for @xprazak2 to finish the proper review that he started :-)

xprazak2:
Test failure is indeed unrelated, migrations no longer fail. I think we are good here. Thanks @shiramax!

PR #296, opened 2017-09-27
Fixes #21124 - Use title when original filename is missing

xprazak2:


ares:
Thanks @xprazak2, works great! Merging.

PR #297, opened 2017-09-27
Fixes #21125 - API call for arf_reports fails for non-admin user

ranjan:


xprazak2:
Works nicely, thanks!

PR #298, opened 2017-10-05
Fixes #21209 - Ensure report digest is present for deleted policies

xprazak2:


ares:
Tests are failing, it seems we wanted to delete the report for deleted policy. Could you check the git log and find what should be the desired behavior?

xprazak2:
Seems like we used to [delete associated reports](https://github.com/theforeman/foreman_openscap/commit/7e81a0451ff9494047ddf1555adec95448537b90#diff-17d0580fad19c49f31bb2c4b1718a263L12) when policy was deleted, I will restore the original behavior.

xprazak2:
Now we delete all associated reports when we delete a policy.

ares:
Ok, that solves it for new deletions. How do we address existing setups where users already deleted policies but still have reports?

ares:
Also I was going to suggest that deleting a policy should show a warning but it already does in confirmation dialog :-) "Delete compliance policy default with all the reports?" so I think this is the right direction.

xprazak2:
I do not see a simple solution on how to fix the existing reports with deleted policy. We would need to fetch report files from proxy and generate digest again or just ask the proxy to do it for us. We have enough information to get to report files even without digest, but we would need to add an endpoint to the proxy just for this purpose. It seems to me it just is not worth it since reports can easily be generated. Unless someone has a large number of 'valuable reports' with deleted policies.

ares:
@xprazak2 could we just detect such reports and delete them in a migration? they should have been deleted anyway

xprazak2:
I added a migration to delete reports without an association to a policy.

xprazak2:
Test failure unrelated, environment data not found in the schema when running bundler.

db/migrate/20171011134112_remove_arf_reports_without_policy.rb
```diff
@@ -0,0 +1,8 @@
+class RemoveArfReportsWithoutPolicy < ActiveRecord::Migration
+  def up
+    ids_to_keep = ForemanOpenscap::ArfReport.unscoped.all.joins(:policy_arf_report).pluck(:id)
+    ForemanOpenscap::ArfReport.unscoped.where.not(:id => ids_to_keep).find_in_batches do |batch|
```
ares: :dart:

ares:
Failure unrelated, merging, thanks @xprazak2!

PR #308, opened 2017-11-03
Fixes #21091 - Do not create Arf Reports without proxy id

xprazak2:


app/models/foreman_openscap/arf_report.rb
```diff
@@ -4,6 +4,7 @@ module ForemanOpenscap
   class ArfReport < ::Report
     include Taxonomix
     include OpenscapProxyExtensions
+    alias_method :destroy_locally, :destroy
```
ares: why is this required? do we need to skip some destroy callback?

app/controllers/api/v2/compliance/arf_reports_controller.rb
```diff
@@ -84,6 +88,12 @@ def handle_download_error(error)
           render_error 'standard_error', :status => :internal_error, :locals => { :exception => error }
         end
 
+        def no_proxy_for_host(asset)
+          msg = _('Failed to upload Arf Report, no openscap proxy set for host %s') % asset.host.name
+          logger.error msg
+          render :json => { :result => msg, :id => '' }, :status => :unprocessable_entity
```
ares: why the response contain empty id?

lib/tasks/foreman_openscap_tasks.rake
```diff
@@ -56,6 +56,14 @@ namespace :foreman_openscap do
     ForemanOpenscap::MessageCleaner.new.clean
     puts 'Done'
   end
+
+  desc "Delete ArfReports without openscap proxy"
```
ares: s/openscap/OpenSCAP/

lib/tasks/foreman_openscap_tasks.rake
```diff
@@ -56,6 +56,14 @@ namespace :foreman_openscap do
...
+  task :clean_reports_without_proxy => :environment do
```
ares: I suppose we don't want to surprise user which is why we don't delete data automatically through migration, in that case, could you please open a PR against the manual describing how and why to use this?

app/controllers/api/v2/compliance/arf_reports_controller.rb
```diff
@@ -84,6 +88,12 @@ def handle_download_error(error)
           render_error 'standard_error', :status => :internal_error, :locals => { :exception => error }
         end
 
+        def no_proxy_for_host(asset)
+          msg = _('Failed to upload Arf Report, no openscap proxy set for host %s') % asset.host.name
```
ares: s/openscap/OpenSCAP/

ares:
Could you please also point me to the place where we need the proxy for the deletion? I fail to understand, why we can't simply destroy the record in Satellite even without openscap proxy set. Thanks

app/controllers/api/v2/compliance/arf_reports_controller.rb
xprazak2: I suppose I could just omit the id entirely.

app/models/foreman_openscap/arf_report.rb
xprazak2: Not a callback, but we have our own [destroy](https://github.com/theforeman/foreman_openscap/blob/master/app/models/foreman_openscap/arf_report.rb#L189) that calls out to proxy to delete the files there. Since the proxy for reports is not set, we just want to destroy what is in db.

ares: So why alias is prefered instead of adding a detection of existing object into this destroy? It would skip the deletion if it can't be performed.

xprazak2:
We do not need the proxy to destroy the record in db. The only reason we call it is to [delete the report xml](https://github.com/theforeman/smart_proxy_openscap/blob/master/lib/smart_proxy_openscap/openscap_api.rb#L66) it has stored. Currently, if you try to destroy a report and proxy is down, then file deletion on proxy is skipped and record is deleted from db, but we log that we failed to delete from proxy.

We can use the current `destroy`, but we get errors on the rake task output, which are not really errors, because we already know those reports have no proxy. I did not want to remove the logging just for the sake of a rake task, so that is why I went for `alias`.



Anyway, I moved the call to proxy into `:before_destroy => :destroy_from_proxy` and I use `delete` in rake task, so now we do not need the alias.

xprazak2:
I use `as_anonymous_admin` with a block and removed the `@report` variable from test.

ares:
Test failures are related

```
ActionController::UrlGenerationError: No route matches {:action=>"show", :controller=>"api/v2/compliance/arf_reports", :id=>nil}
     (ActionController::UrlGenerationError)
/usr/local/rvm/gems/ruby-2.2.5@test_plugin_pull_request-0/gems/actionpack-4.2.9/lib/action_dispatch/journey/formatter.rb:46    
```

I suppose the other two tests which use @report should be also updated

xprazak2:
Updated and all green now.

ares:
Thanks @xprazak2, merging

PR #309, opened 2017-11-10
Fixes #20106 - Add links to latest policy from widget

xprazak2:


app/views/compliance_hosts/show.html.erb
```diff
@@ -8,7 +8,8 @@
     <% if data.has_data? %>
     <div class="col-md-4">
       <div class="stats-well">
-        <h4 class="header ca"><%= _('%s latest report') % policy.name %></h4>
+        <h4 class="header ca"><%= display_link_if_authorized(_('%s latest report') % policy.name,
```
ares: it would be better to use `link_to_if_authorized` that would keep the text but it would not be clickable

see the difference
0label
1label

app/views/compliance_hosts/show.html.erb
```diff
@@ -8,7 +8,8 @@
...
+                                                              hash_for_arf_report_path(:id => data.latest_report.id)) %></h4>
```
ares: This would keep the link visible even for user that does not have permission for this specific report, resulting in 404 page. It's required to specify auth_object so the authorization is not performed only against generic show arf report action. The complete line that works for me:

```
<h4 class="header ca"><%= link_to_if_authorized(_('%s latest report') % policy.name,
                                                hash_for_arf_report_path(:id => data.latest_report.id).merge(:auth_object => data.latest_report)) %></h4>
```

xprazak2:
Updated as suggested.

ares:
Thanks @xprazak2, merging.

PR #311, opened 2017-11-13
Fixes #20886 - Do not call to proxy when ds content not changed

xprazak2:
The reasoning behind this is that if no new file is uploaded, then there will be no change in profiles and therefore we do not need openscap proxy to be online when we only update other attributes.

### DIFF
--- a/app/lib/proxy_api/available_proxy.rb
+++ b/app/lib/proxy_api/available_proxy.rb
@@ -7,7 +7,8 @@ module ::ProxyAPI
       Net::HTTPBadResponse,
       Net::HTTPHeaderSyntaxError,
       Net::ProtocolError,
-      Timeout::Error
+      Timeout::Error,
+      ProxyAPI::ProxyException
     ].freeze
 
     def initialize(args)
--- a/app/models/concerns/foreman_openscap/data_stream_content.rb
+++ b/app/models/concerns/foreman_openscap/data_stream_content.rb
@@ -10,7 +10,7 @@ module ForemanOpenscap
 
       validates_with ForemanOpenscap::DataStreamValidator
 
-      after_save :create_profiles
+      after_save :create_profiles, :if => lambda { |ds_content| ds_content.scap_file_changed? }
 
       before_validation :redigest, :if => lambda { |ds_content| ds_content.persisted? && ds_content.scap_file_changed? }
       before_destroy ActiveRecord::Base::EnsureNotUsedBy.new(:policies)


ares:
Works well but the redmine issue also mentions, we should raise a proper exception in `#proxy_url` method from DataStreamContent if proxy is not available.

It also looks like even if the proxy is down, we try to connect to it, now only if the file changed. The proxy is considered available if it has openscap feature and has a minimum version but we don't check its status. Should it be fixed too? (separate redmine issue would do)

xprazak2:
The ticket suggests to raise an exception as a way to fix this, however I found out that it may not be needed after all. Proxy being nil (and therefore down) is handled by the [validator](https://github.com/theforeman/foreman_openscap/blob/master/app/validators/foreman_openscap/data_stream_validator.rb#L13), now I am getting proper error messages if there is something wrong with the proxy.

We actually make a request to proxy to find out a [version](https://github.com/theforeman/foreman_openscap/blob/master/app/lib/proxy_api/available_proxy.rb#L17-L31), so if we can get the version it means the proxy is up and if not, we say it is not available. So it seems we are good unless I am missing something.

ares:
well, during testing I saw the behavior that we tried to connect to he proxy even when it was marked as down so I got ugly error message, probably from phase when we're detecting features

scap

the trace follows

```
2017-11-27T08:34:53 26f23be6 [proxy] [D] RestClient.get "https://foreman.example.tst:8443/features", "Accept"=>"application/json", "Accept-Encoding"=>"gzip, deflate", "User-Agent"=>"rest-client/2.0.2 (linux-gnu x86_64) ruby/2.4.0p0", "X-Request-Id"=>"eb61151d014061336507ecab6e85d04754cd01041b3deac43f8fca23f395b2b1"
2017-11-27T08:34:53 26f23be6 [sql] [D]    (0.1ms)  ROLLBACK
2017-11-27T08:34:53 26f23be6 [app] [W] ProxyAPI operation FAILED
 | ProxyAPI::ProxyException: ERF12-2530 [ProxyAPI::ProxyException]: Unable to detect features ([Errno::ECONNREFUSED]: Failed to open TCP connection to foreman.example.tst:8443 (Connection refused - connect(2) for "foreman.example.tst" port 8443)) for proxy https://foreman.example.tst:8443/features
 | /home/ares/Projekty/Zdrojaky/foreman/lib/proxy_api/features.rb:11:in `rescue in features'
 | /home/ares/Projekty/Zdrojaky/foreman/lib/proxy_api/features.rb:9:in `features'
 | /home/ares/Projekty/Zdrojaky/foreman_openscap/app/lib/proxy_api/available_proxy.rb:29:in `has_scap_feature?'
 | /home/ares/Projekty/Zdrojaky/foreman_openscap/app/lib/proxy_api/available_proxy.rb:19:in `available?'
 | /home/ares/Projekty/Zdrojaky/foreman_openscap/app/models/concerns/foreman_openscap/data_stream_content.rb:22:in `block in proxy_url'
 | /home/ares/.rbenv/versions/2.4.0/lib/ruby/gems/2.4.0/gems/activerecord-5.0.6/lib/active_record/relation/delegation.rb:38:in `each'
 | /home/ares/.rbenv/versions/2.4.0/lib/ruby/gems/2.4.0/gems/activerecord-5.0.6/lib/active_record/relation/delegation.rb:38:in `each'
 | /home/ares/.rbenv/versions/2.4.0/lib/ruby/gems/2.4.0/gems/activerecord-5.0.6/lib/active_record/relation/finder_methods.rb:65:in `find'
 | /home/ares/.rbenv/versions/2.4.0/lib/ruby/gems/2.4.0/gems/activerecord-5.0.6/lib/active_record/relation/finder_methods.rb:65:in `find'
 | /home/ares/Projekty/Zdrojaky/foreman_openscap/app/models/concerns/foreman_openscap/data_stream_content.rb:20:in `proxy_url'
 | /home/ares/Projekty/Zdrojaky/foreman_openscap/app/validators/foreman_openscap/data_stream_validator.rb:13:in `validate
```

the issue is, we don't get HTTP error here, but rather `Errno::ECONNREFUSED`. Could you please add it to the list of expected failures in `available?`. Then it should work fine.

ares:
Sorry, it's actually ProxyAPI::ProxyException that is raised.

xprazak2:
Updated.

ares:
Thanks @xprazak2, merging.

PR #316, opened 2017-12-01
Fixes #21839 - Pull in the latest strings

bkearney:


ares:
Test failures unrelated, thanks @bkearney, do you want to bring this also to a stable 0.7 branch that is available for Foreman 1.15?

bkearney:
@ares I will leave that up to you. There are some benefits to that downstream, but I will eventually need to add new strings there so it may be better to keep this in master upstream.

ares:
Ok, thanks @bkearney, merging.

PR #319, opened 2018-01-04
Fixes #22146 - mark params coming from proxy as safe

ares:
@xprazak2 could you take a look pls? this is 0.9 blocker


xprazak2:
Ack.

PR #324, opened 2018-01-25
Fixes #22405 - Sort ArfReports by policy and openscap proxy

xprazak2:


ares:
Thanks @xprazak2, fixes the problem indeed.

san7ket:
Works as expected.

scap

PR #325, opened 2018-02-06
Fixes #22521 - add hover effect to tables

ares:


xprazak2:
Works nicely, merging. Thanks!

PR #327, opened 2018-02-26
Fixes #22668 - Always delete report records

xprazak2:


ares:
Does this mean that when user will try to delete it from UI and proxy is down, we'll get into inconsistent state? Could we fix this through foreman-maintain later? If this was found as upgrade issue and we stop deleting reports on proxy which is down, I guess the upgrade can leave leftovers on the proxy?

xprazak2:
I guess we can keep the logic as it is since this bug affects only reports without policy, which is what the [migration](https://github.com/theforeman/foreman_openscap/blob/master/db/migrate/20171011134112_remove_arf_reports_without_policy.rb) is trying to remove. 

So we delete the reports without policy just from the db (and we could not delete them even if proxy was online - we need the digest which is stored in policy_arf_report association object, which is likely missing as well)

I would definitely like to have something that would allow us to clean up directories on proxy, I created [a ticket in forman_maintain](https://projects.theforeman.org/issues/22687).

app/models/foreman_openscap/arf_report.rb
```diff
@@ -189,18 +189,22 @@ def equal?(other)
...
-        end
+      if !host
+        destroy_from_proxy_warning "host"
+      elsif !policy
+        destroy_from_proxy_warning "policy"
+      elsif !openscap_proxy
+        destroy_from_proxy_warning "OpenSCAP proxy"
       else
-        logger.error "Failed to delete report with id #{id} from proxy, no host associated with report"
+        openscap_proxy_api.destroy_report(self, ForemanOpenscap::Helper::find_name_or_uuid_by_host(host))
```
ares: we're no longer rescuing from errors on proxy? does that mean we'd render ugly error if proxy returns 500 e.g.? previously, the method in case of proxy error returned true (because of `logger.debug`), should we perhaps keep the rescue in this else branch?

ares: ah scratch that, I see the rescue is handled in `#destroy_report` already and it returns false

ares:
turns out other error can occure because of taxnomies, the migration should be wrapped in `User.as_anonymous_admin` block, otherwise finding the openscap proxy can fail as User.current is nil during the migration

xprazak2:
I wrapped the migration in `User.as_anonymous_admin` block

ares:
Thanks @xprazak2, works as expected and fixes the issue, merging!. I'll cp to 0.7-stable.

PR #329, opened 2018-03-12
Refs #22668 - use admin only on seeded database

ares:


ares:
Reported this morning, we would have caught it if tests were running :-( @rplevka this should fix the issue, @xprazak2 could you take a look? If DB is seeded, the code runs under admin (in seeded DB, there can be ArfReports), otherwise just run under non-admin, there won't be any reports anyway.

xprazak2:
Fixes the issue for me when running all migrations on new db. Mergign, thanks!

PR #331, opened 2018-03-29
Fixes #22815 - Show properly back button in policy wizard

xprazak2:


ares:
Thanks @xprazak2, fixes the problem. I don't expect tests to be broken by this change :-)

PR #335, opened 2018-05-09
Fixes #23544 - Edit multiple host's OpenSCAP Proxy

treydock:


app/controllers/concerns/foreman_openscap/hosts_controller_extensions.rb
```diff
@@ -10,5 +16,43 @@ def process_hostgroup
...
+        redirect_to hosts_path
+      else
+        error _('No OpenSCAP Proxy selected.')
+        redirect_to(update_multiple_openscap_proxy_hosts_path)
+      end
+    end
+
+    private
+
+    def find_multiple
```
xprazak2: I do not think you need this method. This module extends `HostsController` and there already is [find_multiple](https://github.com/theforeman/foreman/blob/develop/app/controllers/hosts_controller.rb#L725)

xprazak2:
This seems like a cool feature. So far, I am getting: 



Looking at the code, the `update_multiple_openscap_proxy` handles the actual update, but it gets called when I want to display the modal form. You will need to add one more controller [action](https://github.com/theforeman/foreman_openscap/blob/master/app/controllers/policies_controller.rb#L76-L76) and rename your partial so that the form displays properly.

app/controllers/concerns/foreman_openscap/hosts_controller_extensions.rb
```diff
@@ -1,5 +1,11 @@
 module ForemanOpenscap
   module HostsControllerExtensions
+    extend ActiveSupport::Concern
+
+    included do
```
xprazak2: The whole module is [prepended, not included](https://github.com/theforeman/foreman_openscap/blob/master/lib/foreman_openscap/engine.rb#L208), so I do not think this will work.

treydock: @xprazak2 Any suggestions on how to use the before_action in a module that is prepended and not included?  I think I also have to add `define_action_permission` because if I put this in `select_multiple_openscap_proxy`:

```
    def select_multiple_openscap_proxy
      @hosts = find_multiple
      @hosts
    end
```

I end up with this error:

```
2018-05-15 12:26:22 79b673c6 [app] [W] Something went wrong while selecting hosts - ERF42-5531 [Foreman::Exception]: unknown permission for hosts#select_multiple_openscap_proxy
 | Foreman::Exception: ERF42-5531 [Foreman::Exception]: unknown permission for hosts#select_multiple_openscap_proxy
 | /usr/share/foreman/app/controllers/application_controller.rb:151:in `action_permission'
 | /usr/share/foreman/app/controllers/concerns/foreman/controller/action_permission_dsl.rb:19:in `action_permission'
 | /usr/share/foreman/app/controllers/concerns/application_shared.rb:22:in `current_permission'
 | /usr/share/foreman/app/controllers/hosts_controller.rb:660:in `resource_base'
 | /usr/share/foreman/app/controllers/hosts_controller.rb:742:in `find_multiple'
```

And the modal form ends up with the contents of hosts index view.

xprazak2:
I think it can be done by adding `select_multiple_openscap_proxy` into [MULTIPLE_ACTIONS](https://github.com/theforeman/foreman/blob/develop/app/controllers/hosts_controller.rb#L17) when you change `self.prepended` like:

```ruby
def self.prepended(base)
    base::AJAX_REQUESTS << 'openscap_proxy_changed'
    base::MULTIPLE_ACTIONS << 'select_multiple_openscap_proxy'
end
```

but if that does not work, you can always do just 

```ruby
def select_multiple_openscap_proxy
   find_multiple
end
```

Regarding the permissions, something like the following might help:

```ruby
def action_permission
   case params[:action]
   when 'select_multiple_openscap_proxy'
       :view
    else
        super
    end
end
```

treydock:
I think things should be working now.  Tested with a test 1.17 install.

app/controllers/concerns/foreman_openscap/hosts_controller_extensions.rb
```diff
@@ -10,5 +10,38 @@ def process_hostgroup
...
+        error _('No OpenSCAP Proxy selected.')
+        redirect_to(select_multiple_openscap_proxy_hosts_path)
+      end
+    end
+
+    private
+
+    def action_permission
+      case params[:action]
+      when 'select_multiple_openscap_proxy'
```
xprazak2: in `engine.rb`, this one is under `:edit_hosts`, so maybe we should keep it consistent?

xprazak2:
It works nicely now, I just left one comment inline.

app/controllers/concerns/foreman_openscap/hosts_controller_extensions.rb
treydock: Switched to using `:edit` for consistency.

xprazak2: Thanks, you can actually branch on multiple conditions, could you get rid of one branch by changing it to the following please? My last comment, cross my heart :)

```ruby
def action_permission
    case params[:action]
    when 'select_multiple_openscap_proxy', 'update_multiple_openscap_proxy'
      :edit
    else
      super
    end
  end
```

treydock: Change made

xprazak2:
Merging, thanks!

treydock:
@xprazak2 Thanks.  If this could make it into the 1.17 repos that would be helpful.

ares:
Thanks @treydock, would you be up for presenting this nice addition at our community demo?

treydock:
@ares What all is involved in presenting at a community demo?  I've been in the Foreman community since the 0.x days and must admit I've never attended a demo.

ares:
All you need to do is fill in the table at https://community.theforeman.org/t/foreman-community-demo-45/9698 and join the google hangout at the right time. The link for presenter hangout is shared 15 minutes before the demo on theforeman-dev irc channel. Next community demo is tomorrow :-) Few tips about demoing can be found at http://projects.theforeman.org/projects/foreman/wiki/Current_Sprint_Information 

Usually it's good to explain a bit of a context, like why $change was needed, and then show the change in your browser. If you need any help with preparations, please reach out to me (mhulan on IRC) or @GregSutcliffe - gwmngilfen on IRC.

treydock:
@ares What about next week or whenever next demo takes place?  My test environment where I could demo the changes is currently reverted to 1.16.1 to match my production install and would likely not be running 1.17.1 in time for tomorrow's demo.

ares:
No worries, one demo is today and we have 3 weeks cadence, so next chance is 3 weeks from now. Watch  https://community.theforeman.org/c/events/l/agenda or even better this calendar - https://community.theforeman.org/c/events/l/calendar

PR #339, opened 2018-07-13
Fixes #24248 - make job template available

ares:
the same problem we've seen with REX, ansible after 1.15 change how taxonomies work with any context


ares:
@xprazak2 mind to take a like before Thursday? I'd like to get that into another foreman_openscap build.

xprazak2:
Takes care of the issue for me, thanks!

PR #340, opened 2018-07-18
Fixes #24246 - Remove lookup keys extensions

xprazak2:
Hiding the lookup keys in Parameters tab was introduced
so that users would not modify the port, server and policies
keys, which would lead to misconfigureations. But users can
still do that by creating a matcher when editing puppet
classes.

Several new class params were added
and now that we have fetch-remote-resources, I see a value
of being able to change the params directly when editing host.


app/models/foreman_openscap/policy_revision.rb
```diff
@@ -1,6 +0,0 @@
-module ForemanOpenscap
-  class PolicyRevision < ApplicationRecord
-    belongs_to :policy
-    belongs_to :scap_content
```
ares: is this related? is there a leftover table that we should drop?

xprazak2: You are right, migration will be [needed](https://github.com/theforeman/foreman_openscap/blob/master/db/migrate/20141104171545_create_scaptimony_policy_revisions.rb). I will raise a separate issue.

ares:
This requires yet unreleased puppet-foreman_scap_client, so I won't merge right away. Technically it would do any harm merging even without, but the main reason - fetch remote resources - is not there yet.

ares:
[test foreman_openscap]

PR #345, opened 2018-08-01
Fixes #24472 - Identify report source proxy when using LB

xprazak2:


xprazak2:
Needs [proxy part](https://github.com/theforeman/smart_proxy_openscap/pull/60) and there are changes for [installer](https://github.com/theforeman/puppet-foreman_proxy/pull/443) as well.

xprazak2:
[test foreman_openscap]

app/controllers/api/v2/compliance/arf_reports_controller.rb
```diff
@@ -84,12 +80,27 @@ def find_resource
...
         end
 
+        def find_resources_before_create
+          @asset = ForemanOpenscap::Helper::get_asset(params[:cname], params[:policy_id])
+          if !params[:openscap_proxy_url] && !params[:openscap_proxy_name]
+            msg = _('Failed to upload Arf Report, OpenSCAP proxy name or url not found in params for when uploading report for %s') % @asset.host.name
+            no_proxy_for_host(msg)
+            return
+          end
+          @smart_proxy = SmartProxy.find_by :name => params[:openscap_proxy_name]
```
ares: how do we set current org/loc here? should we use `.unscoped` to always find a proxy, context regardless?

app/controllers/api/v2/compliance/arf_reports_controller.rb
```diff
@@ -84,12 +80,27 @@ def find_resource
...
+          @smart_proxy ||= SmartProxy.find_by :url => params[:openscap_proxy_url]
```
ares: the same here

app/controllers/api/v2/compliance/arf_reports_controller.rb
```diff
@@ -84,12 +80,27 @@ def find_resource
           instance_variable_set("@arf_report", resource_scope.find(params[:id]))
         end
 
+        def find_resources_before_create
+          @asset = ForemanOpenscap::Helper::get_asset(params[:cname], params[:policy_id])
+          if !params[:openscap_proxy_url] && !params[:openscap_proxy_name]
```
ares: this now became hard failure, meaning older proxies without https://github.com/theforeman/smart_proxy_openscap/pull/60 won't be able to upload arf reports, correct? should we rather fallback to asset's host proxy in such case to remain compatible, just with the limitation for LB?

xprazak2:
I do not think this affect templates seeding, so test failures probably unrelated.

ares:
Thanks, works fine, tested with proxy with and without the patch. Merging!

PR #347, opened 2018-08-08
Fixes #24563 - Fix scoped searches for arf report

xprazak2:


ares:
I actually have problem reproducing this. "last_for = host" on reports page and "compliance_status" on hosts page seems to work fine. Could you share more info how to reproduce the issue? After applying the patch, I haven't observed any change.

ares:
> You are right, the compliance_status for hosts works, but compliance_passed, compliance_failed and compliance_othered for reports do not. I replaced those three with compliance_status so we now have just one query for passed/failed/othered. It might be a bit confusing that we have a query with the same name for hosts and reports, but I did not come up with a better one.

Ok, no problem with the name. I was just missing good reproducing steps so I though, it's the host one. I can confirm this now works correctly.

> "last_for = host" returns empty results for me when there are puppet reports with higher ids than last openscap report. I needed to ask for the correct report type by adding it into WHERE. If you run the /api/v2/compliance/arf_reports_controller_test.rb with the changes from here against master, the "should find last report for hosts" test will fail.

Ok, now I got it and was also able to reproduce. It was not clear it messes up with config reports.

Thanks, merging!

PR #352, opened 2018-08-23
Fixes #24696 - use string for :class_name attributes

iNecas:
Since Rails 5.2, passing class to `:class_name` is not allowed anymore.


iNecas:
Test failures unrelated

xprazak2:
Thanks!

PR #353, opened 2018-08-27
Fixes #24719 - Override params for puppet class

xprazak2:
Since #24355 was introduced in core, the parameters
are no longer marked for override on import, so we need
to do it ourselves.


xprazak2:
I added a test.

ares:
test failures unrelated, thanks @xprazak2, merging!

PR #354, opened 2018-09-05
Fixes #24820 - Recover from duplicates when adding unique index

xprazak2:


db/migrate/20171016125613_add_content_title_unique_constraint.rb
```diff
@@ -0,0 +1,19 @@
...
+    titles = ForemanOpenscap::ScapContent.unscoped.group(:title).count.select { |key, value| value > 1 }.keys
+    titles.each do |title|
+      duplicates = ForemanOpenscap::ScapContent.unscoped.where :title => title
+      say "#{duplicates.count} Scap Contents with duplicate title detected: #{title}"
+      duplicates.each.with_index do |item, index|
+        next if index == 0
+        new_name = item.title + " #{index + 1}"
+        say "Renaming Scap Content #{item.title} with id #{item.id} to #{new_name}"
+        item.title = new_name
+        item.save!
```
ares: This will be prone to errors if in future we add new validations. We can't use the model or we need to skip validations. You can either use `ForemanOpenscap::ScapContent.update` or define fake model.

ares:
failures caused by rails 5.2, we can ignore this for 0.10

ares:
why did the name of the file changed? it's not a big deal as the number is important, just curious

otherwise I can confirm it works fine

xprazak2:
I changed it, it was 20171016125613_add_content_title_unique_constraint.foreman_openscap.rb. I removed the '.foreman_openscap' so that is has a standardized name.

ares:
this was added as suggested on discourse some time ago, if plugins generate migrations using the core rake task, they are tagged and can easily be migrated down in case people are uninstalling the plugin, could you please keep the name? or I can do it on merge if you're off already

ares:
Merged as ac6d5acd847985e86b214ad66cd6b8eea349375d with the original name, thanks @xprazak2!

PR #351, opened 2018-08-15
Fixes #24606 - Fix breadcrumb bar on compliance host page

xprazak2:

PR #355, opened 2018-09-06
Refs #24606 - hide resource switcher on policy page

ares:
@xprazak2 could you please review/merge for 0.10.3 release, this is alternative to #351 so we don't have to rely on too new Foreman, once we release 0.11 I'm all for your fix :-)


xprazak2:
Given that we do not need those searches, the following will work:

```ruby
<%= breadcrumbs(:resource_url => api_hosts_path,
                 :name_field => 'name',
                 :switchable => false,
                 :items => [
                   { :caption => _('Compliance Hosts') },
                   { :caption => (N_("%s compliance reports by policy") % @host.to_label) }
                 ])
%>

```

ares:
oh true, pushed the new version

xprazak2:
Thanks!

PR #356, opened 2018-09-11
Fixes #24883 - Fix failures when client overrides removed manually

xprazak2:


ares:
Thanks @xprazak2, merging!

PR #357, opened 2018-09-11
Fixes #21127 - Skip logging arf-report body

xprazak2:
The logging of response body is skipped for downloading scap content and tailoring file via API. The report logs are filtered out when proxy uploads new report. The only questions is whether we should rename `:logs` to something more specific like `:openscap_logs` to make sure the newly added filter does not get applied for other requests as well.


ares:
Thanks @xprazak2, merging!

PR #359, opened 2018-09-25
Fixes #25028 - Fix scope lookup for default taxonomies

xprazak2:


ares:
Tests seems to be failing

xprazak2:
The test failures could be related to Rails 5.2 since they are in master as well. I made a change to fix it here.

xprazak2:
[test foreman_openscap]

PR #361, opened 2018-10-09
Fixes #25151 - Properly show spool error

xprazak2:


xprazak2:
Needs https://github.com/theforeman/smart_proxy_openscap/pull/65

ares:
[test foreman_openscap] not sure if this helps, might need a rebase

ares:
needs small rebase now again

app/controllers/openscap_proxies_controller.rb
```diff
@@ -2,7 +2,7 @@ class OpenscapProxiesController < ApplicationController
   before_action :find_proxy, :only => [:openscap_spool]
 
   def openscap_spool
-    last_error = @smart_proxy ? find_spool_error : nil
+    last_error = @smart_proxy ? @smart_proxy.statuses[:openscap].spool_status : nil
```
ares: could we rename `last_errors` to `spool_errors` or just `error` please? that we we don't have to rename in future when we add more error types :-)

app/views/smart_proxies/_openscap_spool.html.erb
```diff
@@ -1,11 +1,11 @@
 <span>
-  <% if last_error && last_error.has_key?('timestamp') %>
+  <% if last_error && last_error['errors_count'] > 0 %>
```
ares: if last_error is renamed, it needs to be renamed everywhere here too

PR #362, opened 2018-11-05
Fixes #25339 - Fix after_save callbacks

xprazak2:


ares:
@san7ket mind to test this patch? :+1:  from code perspective

san7ket:
@xprazak2 @ares Can I test this on nightly setup and what would be the steps?
I tried pulling in the changes on my sat-box and restarted the services , do I need to do anything else. I couldn't see that it fixed the issue, 
Will try with develop setup later..

san7ket:
@xprazak2 @ares Can I test this on nightly setup and what would be the steps?
I tried pulling in the changes on my sat-box and restarted the services , do I need to do anything else. I couldn't see that it fixed the issue, 
Will try with develop setup later..

xprazak2:
@san7ket, when testing on production setup, I usually create a git repo in the location where plugin is installed, add all existing files and create initial commit. Then I add a remote and cherry-pick the change.

ares:
here are my steps

```
scl enable tfm 'gem which foreman_openscap'
cd $dir_of_plugin_from_above
# optionally: git init; git add .; git commit -m 'init'
wget https://patch-diff.githubusercontent.com/raw/theforeman/foreman_openscap/pull/362.diff
git apply 362.diff
service httpd restart
```

san7ket:
@ares @xprazak2 , Thanks for the steps and the fix works mostly. There is just one issue , the scap content has to be re-uploaded after applying the fix. Scap content uploaded before the fix, will still have the same issue , even after the fix and profiles won't be shown for them.

xprazak2:
@san7ket, that is correct, the contents will have to be uploaded again. To add the missing profiles for scap contents, we would have to add code that detects them, then sends them to proxy to generate profiles and updates them. I do not think it is worth the effort since deleting and uploading new scap contents is not too difficult.

ares:
If this was not present in released version, I'd be fine without auto detection

ares:
Since this was introduced by Foreman rails 5 bump, only users running 1.18+ who uploaded new content would be affected. In case this will be needed, foreman-maintain should receive the fix. Merging as is, thanks @xprazak2!

PR #365, opened 2018-12-11
Fixes #25710 - Notify when scap feature missing

upadhyeammit:
This adds error as notification to handle condition when there is 'OpenSCAP proxy' but its actually not having Openscap feature, so user will be notified and can unset or correct the 'OpenSCAP proxy'.


app/controllers/concerns/foreman_openscap/hosts_and_hostgroups_extensions.rb
```diff
@@ -0,0 +1,23 @@
+module ForemanOpenscap
+  module HostsAndHostgroupsExtensions
```
xprazak2: The module name suggests that it extends hosts and hostgroups, but it extends the controllers, so it should be `HostsAndHostgroupsControllerExtensions`.

xprazak2:
Works! I left just one comment inline.

app/controllers/concerns/foreman_openscap/hosts_and_hostgroups_extensions.rb
upadhyeammit: Fixed!

xprazak2:
Thanks!

PR #367, opened 2019-01-08
Fixes #23950 - Add deployment option to policy

xprazak2:
This lifts the restriction on Puppet for foreman_openscap and allows Ansible as an alternative. Policy has a new attribute `deploy_by`, which is used to override Puppet class params/Ansible variables behind scenes.

Still wip, foreman_scap_client Ansible role will need additional work and moving to the proper namespace when it is ready.

Even though there is not too much Ansible specific code, I think this gives us a very strong case for testing plugins together.


ares:
@san7ket have you had a chance to test this? I saw thumbs up but not sure how deep you got. I think we have now everything also for ansible based deployment except for packaging

san7ket:
@ares I haven't tried it yet, will give it a try

app/helpers/policies_helper.rb
```diff
@@ -9,6 +9,33 @@ def policy_profile_from_scap_content(policy)
     policy.scap_content_profile.nil? ? "Default" : policy.scap_content_profile.title
   end
 
+  def deploy_by_radios(f, policy)
+    ForemanOpenscap::ConfigNameService.new.configs.map do |tool|
```
ares: options should either have inline help with some explanation what to expect or the tab should have generic help text explaining that puppet requires puppet agent configured and respective puppet class will be assigned to the host/hostgroup, similarly for ansible where user should be warned about SSL cert setup. For manual, user should be told about what are the steps to be taken for successful deployment.

app/views/policies/_form.html.erb
```diff
@@ -5,7 +5,8 @@
   <%= base_errors_for @policy %>
 
   <ul class="nav nav-tabs" data-tabs="tabs">
-    <li class="active"><a href="#primary" data-toggle="tab"><%= _("General") %></a></li>
+    <li class="active"><a href="#deployment" data-toggle="tab"><%= _("Deployment Options") %></a></li>
```
ares: in wizard we call it Client Deployment, if we plan to add more options in future, let's use this but also in the wizard

app/services/foreman_openscap/client_config/ansible.rb
```diff
@@ -0,0 +1,33 @@
...
+      def type
+        :ansible
+      end
+
+      def available?
+        defined?(ForemanAnsible)
+      end
+
+      def help_text
+        "Make sure you have Ansible plugin installed and #{ansible_role_name} Ansible role imported."
```
ares: if this could include html, it would be great to link to import roles page (the same for puppet classes)

app/services/foreman_openscap/lookup_key_overrider.rb
```diff
@@ -0,0 +1,71 @@
...
+      override_required_params config
+    end
+
+    private
+
+    def override_required_params(config)
+      item = config.find_config_item
+
+      unless item
+        err = _("Required #{config.msg_name} %{class} was not found, please ensure it is imported first.") % { :class => config.config_item_name }
```
ares: #{config.msg_name} is missing translation and should be injected using %

app/services/foreman_openscap/lookup_key_overrider.rb
```diff
@@ -0,0 +1,71 @@
...
+      return unless @policy.deploy_by && Policy.deploy_by_variants.include?(@policy.deploy_by)
+      config = @name_service.config_for @policy.deploy_by.to_sym
+      unless config.available?
+        @policy.errors[:deploy_by] <<
+          _("%{type} was selected to deploy policy to clients, but %{type} is not available. Are you missing a plugin?") %
+          { :type => config.type.to_s.camelize }
+        return
+      end
+      return unless config.managed_overrides?
+      override_required_params config
```
ares: Interestingly, I was able to create policy with ansible deployment with only role imported, no variables were imported, I'd expect that also fails.

ares: with variables imported, it marked them for override but I didn't see the default value or matcher being set for port and server, it only set the erb for policies

san7ket:
What @ares says, on variable import there were no default values, I still went ahead with execution of the role and the foreman_scap_client.yaml on client has all those values missing. 
So, I would say the package deployment was successfully for me , but it didn't configure it.

san7ket:
@xprazak2 I have a suggestion, 
When we have no foreman_scap_client role in foreman, when we create the policy, it would show user an error, that it is missing the role. Can we have something similar if the user has forgotten to import  or missing the required ansible variables in foreman.
Also, another issue, if you have not imported the ansible variable and you try to create host with the role assigned, the host page will break, can we handle this exception aswell

san7ket:
@xprazak2 with the steps I was able to deploy the role successfully and get it configured.

1. Import role and variable
2. Create policy 
3. Assign role to host and run it.

I still feel user should be intimated, if he forgots to import variable , when he is creating policy. Because as you said the variables are overridden when the policy is created. 
If, 
1. Import role 
2. Create policy
3. Create host - Fails
4. Imports Variables
5. Create host - Fails (until he overrides the required variables)

app/services/foreman_openscap/lookup_key_overrider.rb
xprazak2: That is the same behaviour as we have for Puppet. There is no default value for port and server - they are based on smart proxy with openscap feature. They get populated once host(group) has openscap proxy.

app/helpers/policies_helper.rb
xprazak2: I added inline notification with link to docs.

xprazak2:
I renamed the policy wizard steps and made sure errors are shown when Ansible variables are not imported + other minor changes.

ares:
note mostly for myself
it seems to be causing regression in REX NoMethodError: undefined method `digest' for nil:NilClass, though it may be something on my side, since I'm testing more PRs together

```
/home/ares/Projekty/Zdrojaky/foreman_openscap/app/models/foreman_openscap/policy.rb:177:in `to_enc'
/home/ares/Projekty/Zdrojaky/foreman_openscap/app/models/concerns/foreman_openscap/host_extensions.rb:84:in `map'
/home/ares/Projekty/Zdrojaky/foreman_openscap/app/models/concerns/foreman_openscap/host_extensions.rb:84:in `policies_enc'
/home/ares/.rbenv/versions/2.4.0/lib/ruby/gems/2.4.0/gems/safemode-1.3.5/lib/safemode/jail.rb:31:in `method_missing'
Unnamed:52:in `bind'
```

san7ket:
ACK

app/services/foreman_openscap/lookup_key_overrider.rb
ares: I don't think we'd force user to fill that in for smart class parameter manually. I tested that on Foreman 1.18 at least, after we create a policy, we add a matcher for hostgroup for both server and port parameter for this hostgroup. But that does not seem to happen for respective ansible variables. They are just set for overriding. It's fine not to set default value, but the matcher for hostgroup="theoneIselected" should be added. Sorry for confusing with default value.

app/helpers/policies_helper.rb
```diff
@@ -20,7 +20,7 @@ def deploy_by_radios(f, policy)
       end
 
       help_block = content_tag(:span, :class => 'help-block help-inline') do |variable|
-        tool.help_text unless tool.available?
+        config_inline_help tool.inline_help unless tool.available?
```
ares: This does not show anything, the condition should probably be using `if`. I think we could go further with the description and also display it more standard way using label help. Please see https://github.com/xprazak2/foreman_openscap/pull/1 and merge if you agree

xprazak2: Seems like we have different ideas about the help. It showed when `tool.available? #=> false`, and it told users what to do to enable it.

xprazak2:
I merged in the help enhancement, the tooltips are shown all the time now. I also kept the banner for now, I think direct link to relevant docs could be useful. Downstream branding can deface the whole banner if the link causes problems.

What are the steps when overrides for hostgroup and port/server are not set? Whey I finish the wizard, they are there. Does your hostgroup have openscap proxy?

ares:
Ok, so my confusion came from the fact I didn't have openscap proxy assigned to the hostgroup. I understand we don't make the situation but we should come up with something that will at least let user know that openscap proxy is not set correctly and client can't be deployed.

Anyway not something that would block this PR. Let's get it in! Thanks @xprazak2, great work!

PR #368, opened 2019-01-09
Fixes #25821 - use string in `add_controller_action_scope`

iNecas:
- [x] https://github.com/theforeman/foreman/pull/6399

Using class causes preliminary evaluation of api docs, so that plugin
updates are not visible in docs.

### DIFF
--- a/lib/foreman_openscap/engine.rb
+++ b/lib/foreman_openscap/engine.rb
@@ -50,7 +50,7 @@ module ForemanOpenscap
 
     initializer 'foreman_openscap.register_plugin', :before => :finisher_hook do |app|
       Foreman::Plugin.register :foreman_openscap do
-        requires_foreman '>= 1.20'
+        requires_foreman '>= 1.21'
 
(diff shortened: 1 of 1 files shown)


lib/foreman_openscap/engine.rb
```diff
@@ -50,7 +50,7 @@ class Engine < ::Rails::Engine
 
     initializer 'foreman_openscap.register_plugin', :before => :finisher_hook do |app|
       Foreman::Plugin.register :foreman_openscap do
-        requires_foreman '>= 1.20'
+        requires_foreman '>= 1.21'
```
iNecas: Assuming https://github.com/theforeman/foreman/pull/6399 will get to 1.21

iNecas:
Foreman dependency has been merged

[test foreman_openscap]

ares:
Failure unrelated, thanks @iNecas, I'll bump the version to 0.12.0, do we need the release for 1.21/nightly?

PR #379, opened 2019-04-23
Fixes #26662 - Fix adding deploy_by on mysql

xprazak2:
I haven't tested, might help on mysql.

### DIFF
--- a/db/migrate/20190103093409_add_deployment_option_to_policy.foreman_openscap.rb
+++ b/db/migrate/20190103093409_add_deployment_option_to_policy.foreman_openscap.rb
@@ -3,7 +3,7 @@ class AddDeploymentOptionToPolicy < ActiveRecord::Migration[5.2]
     add_column :foreman_openscap_policies, :deploy_by, :string
     ForemanOpenscap::Policy.unscoped.in_batches do |batch|
       batch.map do |policy|
-        policy.update_attribute(:deploy_by, 'puppet')
+        policy.update_column(:deploy_by, 'puppet')
       end
     end
     change_column :foreman_openscap_policies, :deploy_by, :string, :null => false


db/migrate/20190103093409_add_deployment_option_to_policy.foreman_openscap.rb
```diff
@@ -6,6 +6,7 @@ def up
         policy.update_attribute(:deploy_by, 'puppet')
       end
     end
+    ForemanOpenscap::Policy.reset_column_information
```
dforste: I tried manually updating this and it failed. I also tried removing the later change_column and the migration ran but did not update the column for any records. After updating the values manually I was able to update the column in the database.

dforste: Is there a reason you don't just do :default => "puppet"?

xprazak2: Mainly because 'puppet' is not meant to be a default value, but rather one of the options. But I may do it eventually if I do not find another solution. Thanks for testing.

xprazak2:
I pushed changes which made the migration pass on mysql for me.

ares:
[test foreman_openscap] I haven't tested on mysql myself, `update_column` skips callbacks which may cause the difference. Looks good from code perspective.

db/migrate/20190103093409_add_deployment_option_to_policy.foreman_openscap.rb
ares: @dforste did you have a chance to test the last version of the patch? does it help in your environment?

dforste: Yes I had the patch work in my environment.

xprazak2:
[test foreman_openscap]

ares:
Thanks @xprazak2 and @dforste, merging!

PR #381, opened 2019-04-26
Fixes #26693 - Switch Result and Severity in Report

johnpmitsch:
Currently the pass/fail for each rule in the report is in the fourth column and the severity is in the left. In our lab, this confused many people who saw the severity red boxes and thought the rule was still failing. The first column showing the result column should clear up this confusion. This commit switches those columns.

This is what it looks like:

hi

### DIFF
--- a/app/views/arf_reports/_output.html.erb
+++ b/app/views/arf_reports/_output.html.erb
@@ -1,22 +1,22 @@
 <table id='report_log' class="<%= table_css_classes %>">
   <thead>
     <tr>
-      <th><%= _("Severity") %></th>
+      <th><%= _("Result") %></th>
       <th><%= _("Message") %></th>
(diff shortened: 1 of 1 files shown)


xprazak2:
Thanks!

ares:
Looks good, I wonder if we should also use different colors for severity than for status. @rohoover any thoughts on this?

Rohoover:
Sorry for the late reply - just saw this...

I actually think the colors are fine, it's the visual representation...

For pass/fail, I would use our standard check or error icon with current colors.

For severity, work has been done in other products to show these visually through a meter of sorts also with color:
Screen Shot 2019-05-21 at 11 38 43 AM

Thoughts?

xprazak2:
https://github.com/theforeman/foreman_openscap/pull/451

PR #382, opened 2019-04-29
Regenerate .rubocop_todo.yml

xprazak2:
This should make master green again.


ares:
Thanks @xprazak2, merging!

PR #383, opened 2019-05-09
Fixes #26768 - Clicking Policy name should go on Edit page

apuntamb:
When user clicks on policy name, will be redirected to the policy edit page and not the policy dashboard page.

### DIFF
--- a/app/views/policies/_list.html.erb
+++ b/app/views/policies/_list.html.erb
@@ -10,7 +10,7 @@
   <% for policy in @policies %>
     <tr>
       <td class="ellipsis">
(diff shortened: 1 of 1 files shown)


ares:
Thansk @apuntamb, since we now remove the link to dashboard, we also need to add it to the drop down of actions. It should probably be a first action here https://github.com/theforeman/foreman_openscap/blob/master/app/views/policies/_list.html.erb#L36 with a label "Dashboard"

apuntamb:
@ares Thanks for the review. I have made the changes. Since we now give Edit action on clicking policy name, so do we need to remove it from actions dropdown?

ares:
yeah, that would probably make the most sense, @san7ket this may have impact in some automation. To edit the policy, user from now on must click on it's name, edit link from drop down will disappear

PR #386, opened 2019-05-09
 Fixes #18998 - UX improvements

ares:
based on https://github.com/theforeman/foreman_openscap/pull/385, implemented all things besides arf report icons for status, that needs more thinking

### DIFF
--- a/app/controllers/policies_controller.rb
+++ b/app/controllers/policies_controller.rb
@@ -11,9 +11,8 @@ class PoliciesController < ApplicationController
   end
 
   def index
-    @policies = resource_base.search_for(params[:search], :order => params[:order])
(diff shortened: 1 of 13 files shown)


xprazak2:
Works nicely, could you add pagination for tailoring files and policies as well?

ares:
Added, please take another look. Also we'll need to update the manual for 1.0 so that the documentation button links to something that exists

xprazak2:
APJ

xprazak2:
@ares, could you create an issue in Redmine for the status icons that are not a part of this fix so that we do not lose track of that item?

ares:
Yeah, opened as https://projects.theforeman.org/issues/26780, thanks for the review

PR #387, opened 2019-05-10
Fixes #26001 - Search for Arf reports by partial policy name

apuntamb:
Eliminating the method `search_by_policy_name` here, was with the intent that ARF reports can still be searched based on the generic `scoped_search`. With this PR, compliance_policy and policy attributes can be searched with partial policy names.
If this is not the ideal way, I can still mark it as WIP. :-)


xprazak2:
Works for me, thanks!

PR #388, opened 2019-05-13
Fixes #26778 - Compliance report default action wrong

apuntamb:
This PR fixes the default action on Compliance Reports page as Full reports and not Delete.


xprazak2:
Thanks!

PR #395, opened 2019-06-04
Fixes #26943 - Fix arf report bulk delete

xprazak2:
Note to self: double check no ES6 syntax was used in js files.


ares:
Thanks @xprazak2, merging!

PR #396, opened 2019-06-21
Fixes #27099 - Add GraphQL queries for tailoring files

xprazak2:


adamruzicka:
We're trying to keep f-openscap lean so I'm a bit hesitant to pull in things we don't strictly need, so Ill go ahead and close this. If a need for this ever comes, we can always revive it.

PR #402, opened 2019-07-18
Fixes #27332 - Speed up the Arf reports loading for host

hao-yu:
Having many reports will dramatically slow down the query to retrieve
reports for a host when Organization.current is set. This triggers the
default scope to filter the reports by Organization. If there are millions
of reports to be filtered, Foreman will spend extremely long time to
construct the report ids to be filtered and also slow down the db query.

This causes the "All Hosts" page to load very slow as Foreman needs to
retrieve the status for all the hosts.

This commit fixed the issue by preventing Foreman to run the default scope.

For more info, you can read the bugzilla or the issue.


xprazak2:
Thanks!

PR #406, opened 2019-08-05
Fixes #27358 - Do not allow mixed policy deployments

xprazak2:


xprazak2:
Current step was empty string for me, so I needed `empty?`.

ares:
that was the case here too, but `steps.index(current_step)` then returned nil, hence `steps.index(current_step) || steps.size` worked

PR #407, opened 2019-08-14
Fixes #27628 - Make template for scans work with ansible

xprazak2:


apollo13:
Hi, if I were to test this by monkeypatching this locally, where would I have to set `policies_enc_raw` so that it's found?

xprazak2:
@apollo13, you do not need to set it, it is a [method](https://github.com/theforeman/foreman_openscap/blob/master/app/models/concerns/foreman_openscap/host_extensions.rb#L103) that collects all policies for a host so that the ids can be used in template.

apollo13:
@ares If it makes any difference, I have applied the patches locally and verified that the outcome is expected. Seems to work fine here.

ares:
Thanks @apollo13 that helps greatly! I'll recheck with puppet and that the input is correctly dropped from the job template, but then we're good to go.

PR #408, opened 2019-08-16
Fixes #27569 - Migrate to newer donut charts

xprazak2:


xprazak2:
@lzap to test you need a host and you can choose if you want to go the Ansible way (with foreman_ansible) or the Puppet way. You will also need a proxy with openscap feature.

1) import [puppet-foreman_scap_client](https://github.com/theforeman/puppet-foreman_scap_client) or [ansible-foreman_scap_client](https://galaxy.ansible.com/theforeman/foreman_scap_client). You will need to import variables for ansible as an additional step.

2) create a policy and assign it with openscap proxy and ansible role/puppet class to the host, host needs to have access to [client repo](http://yum.theforeman.org/client/latest/el7/x86_64/) to install foreman_scap_client

3) run puppet/ansible

4) generate a report on client by running `foreman_scap_client $policy_id`, where policy_id
can be found in /etc/foreman_scap_client/config.yaml

lzap:
Works fine, however the bar chart is a bit off, can you do something with it? Firefox stable:

obrazek

xprazak2:
I adjusted the width for graphs so that they do not overflow their corresponding well.

lzap:
Nipick - can you change the layout of the three bootstrap columns from 4, 4, 4 to something like 5, 5, 2? Because now the last column (the table with details) is formatted down when you zoom in just a little and it does not really need to be that wide.

obrazek

xprazak2:
I adjusted grid columns to 5-5-2

lzap:
Looks great, WFM.

obrazek

I don't have merge permissions tho. @ares would you mind?

ares:
Thanks @xprazak2 and @lzap for the review. I'm fine with merging, just one question, should this bump foreman dependency in engine.rb and major version, given it depends on the core charts change? Or this does not depend on it, only line charts do?

xprazak2:
This does not require any changes in core, Donut and Bar charts are already present, so it is just a matter of using them.

ares:
Thanks! Merging then.

PR #413, opened 2019-09-02
Fixes #27756 - Handle error when puppet class absent

xprazak2:


apollo13:
@xprazak2 So manually running this job-template requires puppet to be enabled and running? How to run this with only the remote_execution plugin against a host that is not puppet managed?

xprazak2:
@apollo13, I am working on that as well: https://github.com/theforeman/foreman_openscap/pull/407

apollo13:
Ah perfect, many thanks!

PR #415, opened 2019-09-05
Fixes #27707 - Migrate to newer line charts

xprazak2:
depends on https://github.com/theforeman/foreman/pull/6999

### DIFF
--- a/app/assets/javascripts/foreman_openscap/scap_hosts_show.js
+++ b/app/assets/javascripts/foreman_openscap/scap_hosts_show.js
@@ -1,4 +1,4 @@
 $(function() {
   // Not sure about this ugly hack.
-  $('.col-md-4 .stats-well').height($('.col-md-8 .stats-well').height() + 28);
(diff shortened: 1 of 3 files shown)


ares:
meaning depends on 1.24, do we need to bump the dependant version in engine.rb? also is the core PR cherry-pickable to older version? asking mostly risk-wise, it's +555 lines of code...

xprazak2:
It is cherry-pickable, I did not bump the version in engine.rb because we need a new release for downstream and I was planning to cherry-pick it there.

ares:
Interesting, I thought I made it mergeable, but now I can't rebase. Also I'm trying to reproduce the charting issue but I can't. I'm no longer able to reproduce the donut chart issue either. Anyway @xprazak2 do recal if the line chart was broken too and user needed to resize the browser? If so, could you please rebase at your first occasion so we can get this in? If not, we can keep this one for next release bumping the Foreman dependency.

PR #416, opened 2019-09-10
Regenerate mo files - i18n

lzap:
From the git history I can see that PO files are correctly downloaded and merged, POT extracted but it's necessary also to keep MO files up to date. This silence many warnings everytime OpenSCAP is required.

I suggest you this helper script so you won't forget:

https://github.com/theforeman/foreman_discovery/blob/develop/update-i18n


lzap:
https://projects.theforeman.org/projects/foreman/wiki/Translating#Generating-gettext-translate-tables

PR #418, opened 2019-09-30
Fixes #27959 - Destroy assets together with host

xprazak2:
I found out this is caused by assets being present after a host is deleted


ares:
Thanks @xprazak2, merging!

PR #419, opened 2019-10-01
Fixes #27936 - Fix changing deploy_by

xprazak2:


ares:
Test failures related

xprazak2:
I know what is going on - this is not really testable without ansible. I can mock in tests that ansible is present, but I have no way of finding out whether ansible role is present, because that model does not exist.

app/controllers/api/v2/compliance/policies_controller.rb
```diff
@@ -69,7 +69,13 @@ def create
       param_group :policy
 
       def update
-        process_response @policy.update_attributes(policy_params)
+        ::ForemanOpenscap::Policy.transaction do
```
ares: could you pleace extract this logic to model's method, somehing like

```ruby
def change_policy_deploy_type(params)
  self.class.transaction do
    ...
  end
end
```

and then reuse it in both controllers? it needs to return true/false and base on that, we'd process_response or process_success/process_error

ares:
Thanks @xprazak2, merging

PR #420, opened 2019-10-17
WIP: Fix "SCAP Content" menu entry

quba42:
* Converts the "SCAP Content" menu entry to title case.
* Converts the uncountable noun "Content" to singular.

Relates to: https://projects.theforeman.org/issues/25223

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ Make sure that
 
 1. Log-in to Web Interface
 2. Create new SCAP Content
-  1. Go to *Hosts -> Compliance -> SCAP contents* page
+  1. Go to *Hosts -> Compliance -> SCAP Content* page
   2. Upload DataSteam file
 3. Create new Policy
   1. Go to Hosts -> Compliance -> Policies page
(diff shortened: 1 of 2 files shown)


quba42:
The reason it is "WIP" is that converting the term "Content" to singular means that certain paths and variables should also be changed. This is more involved and needs more testing, that I have not yet gotten around to...

ares:
Thanks for the patch, while we're on topic, what do you think about extracting the whole menu to first level category?

quba42:
Does "first level category" mean that Open SCAP would have its own menu next to "Monitor", "Content", "Hosts", "Configure", etc. ?
If so, I don't have any particularly strong opinions on that.
I do feel like the menu path **Hosts > Compliance > Reports**, would fit more naturally under **Monitor > Reports > Compliance**.
On the other hand, I am hesitant to spread out the Open SCAP plugins menu entries all over the UI.
Perhaps there needs to be a general discussion whether plugins should keep their menu entries together or whether they should be spread around by topic.
(As another example, I hate the fact that Katello "Content Hosts" are in the "Hosts" menu and not in the "Content" menu, imho it makes for a really awkward workflow.)

quba42:
Perhaps we should really open a discussion on discourse. I feel like my previous comment goes beyond the scope of a PR. :smile:

quba42:
@adamruzicka If I did not link such a discussion here, then no, it probably did not happen.

quba42:
Given this is more than five years old, and I have not found the time to push this further, I am closing this PR.

PR #425, opened 2019-11-14
Fixes #28251 - Do not assign lone taxonomies to report

xprazak2:


tbrisker:
Looking at the code, we should be skipping all reports in core like we do for parameteres

ares:
we include Taxonomix in ArfReport, why do we want to stop assigning lone org/loc here?

tbrisker:
Because reports have taxonomies through hosts, they shouldn't be directly assigned: https://github.com/theforeman/foreman/blob/develop/app/models/report.rb#L16-L17

ares:
Oh, that makes sense. I'll merge this to have a fix in place, but I assume similar patch is needed in core. Thanks @xprazak2!

tbrisker:
a bit late but here is the fix in core: https://github.com/theforeman/foreman/pull/7275/

PR #426, opened 2019-11-14
Fixes #28262 - Properly assign lone taxonomies

xprazak2:


ares:
Why do we want to skip this? Shouldn't policy e.g. be autocreated in org/loc if it's the only one?

xprazak2:
I suppose we could have it assigned to org/loc, but the way current assignment is done does not quite work here and reverting to the original behaviour was the simplest solution.

ares:
Thanks @xprazak2, merging!

PR #429, opened 2019-12-11
Fixes #28481 - Make scap proxy port an int in enc

xprazak2:
Needed for newly added type checks in puppet-foreman_scap_client that enforce port to be an integer instead of a string.

These changes handle new installations, I'll need to create a rake task for existing ones, therefore wip.


xprazak2:
No longer wip.

PR #431, opened 2019-12-17
Fixes #28523 - Add profiles api

xprazak2:


ares:
looks good, but tests were not ran since rubocop fails :-( also we'll probably need hammer patch for this

xprazak2:
Only sqlite is failing now.

PR #432, opened 2020-02-11
Fixes #28982 - Add description to plugin roles

stejskalleos:
PR is part of  [Add default descriptions to all roles added from plugins](https://projects.theforeman.org/issues/28936) task.

### DIFF
--- a/lib/foreman_openscap/engine.rb
+++ b/lib/foreman_openscap/engine.rb
@@ -122,12 +122,14 @@ module ForemanOpenscap
                      :resource_type => 'SmartProxy'
         end
 
-        role "Compliance viewer", %i[view_arf_reports view_policies view_scap_contents view_tailoring_files view_openscap_proxies]
+        role "Compliance viewer", %i[view_arf_reports view_policies view_scap_contents view_tailoring_files view_openscap_proxies],
+             "Role granting read permissions to policy configuration, scan results and downloading reports."
         role "Compliance manager", %i[view_arf_reports view_policies view_scap_contents
                                       destroy_arf_reports edit_policies edit_scap_contents assign_policies
                                       create_policies create_scap_contents destroy_policies destroy_scap_contents
                                       create_tailoring_files view_tailoring_files edit_tailoring_files destroy_tailoring_files
-                                      view_openscap_proxies]
+                                      view_openscap_proxies],
+             "Role granting all permissions to compliance features to non-admin users."
         role "Create ARF report", [:create_arf_reports] # special as only Proxy can create
 
         add_all_permissions_to_default_roles


stejskalleos:
Hi @xprazak2, can you please give me some good description for the roles? Thanks

lib/foreman_openscap/engine.rb
```diff
@@ -121,12 +121,14 @@ class Engine < ::Rails::Engine
                      :resource_type => 'SmartProxy'
         end
 
-        role "Compliance viewer", %i[view_arf_reports view_policies view_scap_contents view_tailoring_files view_openscap_proxies]
+        role "Compliance viewer", %i[view_arf_reports view_policies view_scap_contents view_tailoring_files view_openscap_proxies],
+             "TODO"
```
ares: While we're on this, is there a plugin DSL that automatically create these two roles? Something like `add_default_roles`. That would make sure we always have all available view permissions. Not sure what description it would generate though. I guess person with this role is probably a compliance auditor or supervisor. It allows access to policy configuration and scan results and allows downloading reports from the system.

lib/foreman_openscap/engine.rb
```diff
@@ -121,12 +121,14 @@ class Engine < ::Rails::Engine
...
         role "Compliance manager", %i[view_arf_reports view_policies view_scap_contents
                                       destroy_arf_reports edit_policies edit_scap_contents assign_policies
                                       create_policies create_scap_contents destroy_policies destroy_scap_contents
                                       create_tailoring_files view_tailoring_files edit_tailoring_files destroy_tailoring_files
-                                      view_openscap_proxies]
+                                      view_openscap_proxies],
+             "TODO"
```
ares: This one allows full access to compliance features to non-admin users.

lib/foreman_openscap/engine.rb
stejskalleos: So the role description could be something like this?
`Role granting all permissions to policy configuration, scan results and downloading reports.`

lib/foreman_openscap/engine.rb
stejskalleos: So the role description could be something like this?
`Role granting all permissions to compliance features to non-admin users.`

ares: sounds good to me

lib/foreman_openscap/engine.rb
ares: not `all` but `read` or `view`

stejskalleos:
Updated, @ares ready for review

stejskalleos:
[test foreman_openscap]

xprazak2:
LGTM. Rebasing on the latest master should fix the errors, you can also squash the commits.

stejskalleos:
@xprazak2 done.

xprazak2:
Thanks!

PR #433, opened 2020-02-14
Fixes #29009 - Remove check for compliance status

xprazak2:
Should fix current failures on master.


xprazak2:
[test foreman_openscap]

PR #435, opened 2020-03-08
Fixes #29290 - Update Rubocop to 0.80

mmoll:


mmoll:
[test foreman_openscap]

app/controllers/api/v2/compliance/scap_contents_controller.rb
```diff
@@ -51,7 +51,7 @@ def create
       param_group :scap_content
 
       def update
-        process_response @scap_content.update_attributes(scap_content_params)
```
xprazak2, line 54: Should we wait for Rails 6 with this?

mmoll: It is 100% Rails 5.x compatible (at least in core it was enabled for Rails 5.1).

ezr-ondrej: Yeah, it is actually rails 4 compatible, it's alias ever since (see rails/rails#8705)

xprazak2: Ok, thanks for confirming that.

PR #444, opened 2020-07-14
Fixes #30404 - Add DSL docs for OpenSCAP plugin

ofedoren:
@xprazak2, I'm not 100% sure if those are only updates to Jail/methods that can be used in template writing...


xprazak2:
Thanks!

PR #450, opened 2020-08-07
Refs #30358 - Delete assets only for selected policy

xprazak2:


xprazak2:
Destroying assets uncovered an underlying bug - we need to destroy assets only for the policy that we are trying to unassign.

ares (approved):
Thanks @xprazak2, merging

PR #451, opened 2020-08-10
Fixes #26780 - Add icons for report severity

xprazak2:
report-severity

This adds webpack and react to foreman_openscap.


ares:
looks good, is the webpack integration based on the latest code of foreman template plugin? cc @amirfefer could you please check whethet it follows all core recommendations of how to use react in a plugin?

btw I'm ok with RuleSeverity component living in here for now, however I know insights, openscap and leapp plugins are using that already. Perhaps something we should put in core later?

Otherwise looks good. There's more we could enahnce on this table, perhaps we could turn that to the list view we use on audits page at some point, but that's out of the scope for this PR :-)

xprazak2:
This should be in line with recommendations on using react in plugins. I have plans on improving the whole page but I left this intentionally small - main purpose here was to have a small improvement that would introduce the new js stack into foreman_openscap.

ares:
Sounds good, I'll go ahead and test this, code ACK would be nice, but if we won't get in soon, I'm willing to merge.

ares:
This is ready, however would complicate using foreman_openscap with older version of Foreman, keeping open for now after a quick chat.

xprazak2:
I think we can go ahead with this, I think we will not be doing any backports to older foreman versions.

ares:
Nice, just the icons size seems really big, see the screenshot
icons

good to merge once resolved

ares:
Ah, the size seems to change based on the width of the table/window. I guess it shoud rather be fixed.

xprazak2:
Interesting, the image was scaled down for me as per screenshot above. I tweaked css a little, I hope it helps.

ares:
I can still reproduce, not sure if it can be my environment only. However I see it in chrome 89 and FF 86

xprazak2:
Ah, chrome upgrade helped me to reproduce, it should be fixed now.

ares (approved):
Thanks @xprazak2, works fine now! Merging!

PR #452, opened 2020-10-07
Fixes #31008 - add status_link method to compliance-status

yifatmakias:
Compliance status on hosts/show page should be clickable and lead to the last report for the host.
In order to add this in foreman there is a need to create a status_link method in the compliance status.

### DIFF
--- a/app/models/foreman_openscap/compliance_status.rb
+++ b/app/models/foreman_openscap/compliance_status.rb
@@ -8,6 +8,10 @@ module ForemanOpenscap
       N_('Compliance')
     end
 
+    def status_link
+      host.arf_reports_path(:search => "host = #{host.name}")
+    end
+
     def self.bit_mask(status)
       "#{ArfReport::BIT_NUM * ArfReport::METRIC.index(status)} & #{ArfReport::MAX}"
     end


yifatmakias:
related to this PR:
https://github.com/theforeman/foreman/pull/8044

app/models/foreman_openscap/compliance_status.rb
```diff
@@ -8,6 +8,10 @@ def self.status_name
       N_('Compliance')
     end
 
+    def status_link
+      host.arf_reports_path(:search => "host = #{host.id}")
```
xprazak2, line 12: Should the search be `host = #{host.name}`?

xprazak2:
One minor nit, otherwise looks good.

xprazak2:
Works well together with the changes in core.

app/models/foreman_openscap/compliance_status.rb
ares: ir should it even link to the last report directly? note that status represents results of the last scan

xprazak2: If host has multiple policies and several of them are failing, showing just one report might be misleading. I think we are better off showing a list of reports rather that a single one.

xprazak2:
It great that foreman_openscap has another contributor, thanks!

ares:
@yifatmakias great addition, would you mind demoing it? Please sign up at https://community.theforeman.org/t/foreman-community-demo-87/21229, people should see how they can improve their plugins in the same way :-)

yifatmakias:
@ares I think it is a good idea to demo it, I will sign up for the next demo. Thanks :)

PR #453, opened 2020-11-01
Fixes #31186 - Removed method name collision with katello

ShimShtein:
Since the `inherited_ancestry_attribute` is used only once, removed it so it won't collide with Katello's private method.

Also removed method duplication from `proxy_core_extensions`, since the method is not relevant for hosts (the original `inherited_ancestry_attribute` is not defined on host) and already added to hostgroup by `hostgroup_extensions`.

### DIFF
--- a/app/models/concerns/foreman_openscap/openscap_proxy_core_extensions.rb
+++ b/app/models/concerns/foreman_openscap/openscap_proxy_core_extensions.rb
@@ -33,10 +33,6 @@ module ForemanOpenscap
       end
     end
 
-    def inherited_openscap_proxy_id
-      inherited_ancestry_attribute(:openscap_proxy_id)
-    end
-
     private
 
     def scap_client_lookup_values_for(lookup_keys, model_match)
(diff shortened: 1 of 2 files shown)


ShimShtein:
@xprazak2, IMHO test failures are unrelated - they are about a `digest` method.

app/models/concerns/foreman_openscap/openscap_proxy_core_extensions.rb
```diff
@@ -33,10 +33,6 @@ def update_client_params(model_match, config)
       end
     end
 
-    def inherited_openscap_proxy_id
-      inherited_ancestry_attribute(:openscap_proxy_id)
-    end
```
tbrisker, line 38: Why is this one removed? shouldn't it be similar to the hostgroup one?

ShimShtein: > Also removed method duplication from proxy_core_extensions, since the method is not relevant for hosts (the original inherited_ancestry_attribute is not defined on host) and already added to hostgroup by hostgroup_extensions.

@tbrisker, as I have mentioned earlier ^^^. I don't think the method was working for hosts and we weren't using it anywhere.

tbrisker: oops, my bad.

tbrisker (approved):

PR #454, opened 2020-11-02
Fixes #31230 - Do not use digest in messages

xprazak2:
Digest for report messages was dropped as a part
of #30820, so we should no longer use it.


tbrisker (approved):
LGTM, let's see that tests pass

ares (approved):
Thanks @xprazak2, merging!

PR #456, opened 2020-11-05
Fixes #31034 - Add initial OVAL policy

xprazak2:
Adds a model and API controller
for OVAL policy


xprazak2:
Closing in favour of #458

PR #457, opened 2020-11-06
Fixes #31033 - Add model for OVAL content

xprazak2:
Adds initial model and API controller
for OVAL content


xprazak2:
It would be good to add a fixture of the actual xml file once we settle on which files to use.

xprazak2:
Closing in favour of #458

PR #459, opened 2020-11-24
Fixes #31381 - Register REX features in to_prepare block

adamruzicka:
When openscap tried to register REX features, REX was not registered as a plugin
yet and because it was not registered yet, the feature registration got skipped.

This had two causes, features being registered from within a plugin registration
block and foreman_openscap being registered before remote execution as a plugin.
To_prepare block is run after all the plugin-registering initializers so it
addresses both causes.

### DIFF
--- a/lib/foreman_openscap/engine.rb
+++ b/lib/foreman_openscap/engine.rb
@@ -180,19 +180,6 @@ module ForemanOpenscap
                         :description => proxy_description,
                         :api_description => N_('ID of OpenSCAP Proxy')
 
-        if ForemanOpenscap.with_remote_execution?
-          options = {
-            :description => N_("Run OpenSCAP scan"),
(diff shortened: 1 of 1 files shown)


lzap:
It does work to fix the problem:

```
[root@rc foreman_openscap-4.1.0]# wget https://github.com/theforeman/foreman_openscap/pull/459.patch

[root@rc foreman_openscap-4.1.0]# patch -p1 < 459.patch 
patching file lib/foreman_openscap/engine.rb

[root@rc foreman_openscap-4.1.0]# foreman-rake db:seed
Apipie cache enabled but not present yet. Run apipie:cache rake task to speed up API calls.
User with login admin already exists, not seeding as admin.
```

xprazak2:
Thanks everyone!

PR #460, opened 2020-12-01
Fixes #31423 - Update port param type

xprazak2:


ares:
should we rather call save! to know if this failed? also what if the migration has been already applied? the bug does not seem to be triggered during the migration

xprazak2:
I added a new migration to make sure it runs.

ares (approved):
Thanks @xprazak2, merging!

ares:
the tests fail on rubocop

PR #462, opened 2020-12-03
Fixes #31423 - Update port param type

xprazak2:
Cherry-pick for 4.0-stable


xprazak2:
The tests are failing because core removed the digest field.

PR #466, opened 2021-01-18
Fixes #27378 - make deploy options clickable

ares:


ares:
This makes the label clickable and selecting the option, but keeping the little popover icon just displaying the help text without selecting the radio.

xprazak2:
Works, the tooltip help is moved now, do we need an additional styling?

tooltip

ares:
The reason was I had too small width :-) I added float and also adjusted the width so it's consistent with how we use it elsewhere. Ready for rereview.

xprazak2:
Thanks!

PR #467, opened 2021-02-11
Fixes #30970 - Replace mount_react_component helper with react_component

xprazak2:


ares:
I've tested what I could think of. This introduces one regression, if I enter the policy dashboard, the Host Breakdown Chart is not rendered. Without this patch, it shows up. Funny thing is, the dashboard widget (compliance reports breakdown) works fine.

After fixing, good to merge

xprazak2:
Should be fixed now, the cause was that the donut widget expects data with `:data` key in hash.

ares (approved):
Works fine now, merging! Thanks @xprazak2

PR #468, opened 2021-03-02
Fixes #31986 - Update puppet port without override

xprazak2:


ares (approved):
Thanks @xprazak2, merging

PR #472, opened 2021-03-21
Fixes #32147 - move ARF Reports from Hosts menu to Monitor

MariaAga:
As was suggested in the menu rename discussion https://docs.google.com/document/d/1eFbHZjuZC40SYqAn93dOXU0nUf3RnL10Bb8uHIk60w4/edit?usp=sharing
Related to: https://github.com/theforeman/foreman/pull/7809
I'm not sure it should be merged before or after the change

### DIFF
--- a/lib/foreman_openscap/engine.rb
+++ b/lib/foreman_openscap/engine.rb
@@ -142,9 +142,10 @@ module ForemanOpenscap
         menu :top_menu, :compliance_contents, :caption => N_('SCAP contents'),
                                               :url_hash => { :controller => :scap_contents, :action => :index },
                                               :parent => :hosts_menu
-        menu :top_menu, :compliance_reports, :caption => N_('Reports'),
+        menu :top_menu, :compliance_reports, :caption => N_('ARF Reports'),
                                              :url_hash => { :controller => :arf_reports, :action => :index },
-                                             :parent => :hosts_menu
+                                             :parent => :monitor_menu,
+                                             :after => :report_templates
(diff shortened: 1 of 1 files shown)


xprazak2:
I do see logic behind moving this but I am not entirely convinced about the benefits. As a user, I would probably ask: "Where did my reports go?" since it is not obvious it has been moved. Also, openscap related items would be scattered across 2 different menus.

MariaAga:
Tagging @ares, as he suggested the change.

ares:
@MariSvirik any thougths from UXD perspective? Today the menu structure is

Hosts -> Compliance (cateogory) which contains all compliance related menu items, one of them is link to reports. However this change moves Hosts -> ARF Reports (from Compliance group) to Monitor -> Reports (category). You can see that in the demo hw (the version before move).

The concern here is, we are moving something people are probably used to. On the other hand, it seems as a better place, since all reports should be in the reports category. What would you prefer?

MariSvirik:
@MariaAga @ares 
I would suggest moving it to Reports. 
But we should inform a user about this change somehow. Would it make sense to use an inline notification on another page within the Compliance group that is connected to reports? 
Or do we have any other way how we can inform users about changes? CRHC uses Pendo and Resource center to do that.

MariSvirik:
If there are more changes to come let's all do all of them together. After that, I can write a blog article about changes in navigation - that's going to be another way how to introduce them to the users. 

Rox pointed out we have smth like a notification drawer (a bell icon) in Satellite where we can store information about changes as well.

PR #473, opened 2021-04-21
Switch to gh actions for tests

xprazak2:
We need to test with foreman_ansible
to ensure that logic responsible
for config deployment to client
is tested properly. Our Jenkins currently
does not offer any support for testing
multiple plugins together, so a completely
new job would have to be created in JJB.
Swithching to gh actions is easier
and does not require changes
in a different repo.


ares:
Do you want to still test anything?

ares:
Ok after a quick chat we're ready, thanks @xprazak2, merging!

PR #476, opened 2021-05-05
Fixes #32525 - Minor fixes for OVAL Content API

xprazak2:
* replace 'title' with 'name' in API doc
* add url attribute to response
* add API doc for search and pagination


ares (approved):
Thanks @xprazak2, merging!

PR #477, opened 2021-05-10
Fixes #32462 - Add OVAL Policies page

xprazak2:


ares:
Displayes the list of policies, thanks @xprazak2, merging. The errors seems to be the same as previously.

PR #478, opened 2021-05-11
Fixes #32573 - Add OVAL Policy show page

xprazak2:


ares (approved):
Nice, works pretty well! There seems to be some issue with the hosts count in the CVE list, it tells me it's found on two hosts, while I scanned just one and the search also shows me just one. But it could be something with my DB that got corrupted during testing, we'll find out later.

PR #479, opened 2021-05-11
Fixes #32579 - Add OVAL scan template and rex feature

xprazak2:


app/views/job_templates/run_oval_scans.erb
```diff
@@ -0,0 +1,24 @@
...
+feature: foreman_openscap_run_oval_scans
+provider_type: SSH
+snippet: false
+provider_type: SSH
+kind: job_template
+template_inputs:
+- name: oval_policies
+  description: Comma separated OVAL Policy Ids to run
+  input_type: user
+%>
```
ares, line 14: it would be good to have `-` in here to avoid new line

ares (changes requested):
The REX feature is added, however the template is not assigned, therefore it fails when I try to run the scan. This could be my instance being broken (seeding issue I need to resolve). But also I realized this will show up regardless of lab features enabled. I think it will be better to postpone this until the UI is no longer experimental, as this would cause confusion on people who didn't enable the feature. Could we set the `oval_options[:host_action_button]` to `true` only if the setting is enabled? I know it would require Foreman restart to change, but I think we don't have a better option.

Leaving a minor comment, otherwise looks good.

xprazak2:
The button should be hidden, I made changes to the template and hopefully the tests will pass now.
I am not sure what could cause the problems with template seeding, when I deleted the existing one and re-run seed, it was created for me.

ares (approved):
Thanks @xprazak2, merging!

tbrisker:
Looks like this somehow broke auto reloading in development, i'm getting  
```
PG::NotNullViolation: ERROR: null value in column "host_action_button" violates not-null constraint DETAIL: Failing row contains (9, foreman_openscap_run_oval_scans, Run OVAL scan, Run OVAL scan, , 276, null, null, null). 
/home/tbrisker/gits/foreman_remote_execution/app/models/remote_execution_feature.rb:61:in `block in register'
/home/tbrisker/gits/foreman_remote_execution/app/models/remote_execution_feature.rb:56:in `register'
/home/tbrisker/gits/foreman_openscap/lib/foreman_openscap/engine.rb:290:in `block in <class:Engine>'
lib/foreman/middleware/logging_context_request.rb:11:in `call'

xprazak2:
I have been getting that as well, I think this error might be related to the disappearing settings bug since `oval_options[:host_action_button] = Setting[:lab_features]`

PR #480, opened 2021-05-12
Fixes #32587 - Scan all hostgroups for OVAL policy

xprazak2:
Needs #479, therefore wip


xprazak2:
Runs scan on all hosts from a hostgoup and also on hosts in descendant hostgroups as they have inherited the policy.

ares (approved):

ares:
Thanks @xprazak2, merging! Tests were successful before the rebase, no need to wait.

PR #483, opened 2021-06-04
Fixes #32719 - Add new OVAL policy page

xprazak2:
The minimal version of wizard that allows creating an OVAL policy. Needs cleanup, tests and handling permissions. Opening mostly for visibility so we can decide how the individual steps should look like.

General step has schedule only by cron for now, we can add other types later:
steps-general

Hostgroup selection is a second step for now. Do we want table with checkboxes? Dual select? Something else? 
steps-hostgroups

Thoughts about other steps? Selecting hosts?

### DIFF
--- a/app/graphql/mutations/oval_policies/create.rb
+++ b/app/graphql/mutations/oval_policies/create.rb
@@ -1,0 +1,33 @@
+module Mutations
+  module OvalPolicies
+    class Create < ::Mutations::BaseMutation
+      description 'Creates a new OVAL Policy'
+      graphql_name 'CreateOvalPolicyMutation'
+
+      resource_class ::ForemanOpenscap::OvalPolicy
+
+      argument :name, String
+      argument :description, String, required: false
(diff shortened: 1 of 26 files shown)


xprazak2:
I made changes to shrink wizard into a simple form. No longer wip, needs https://github.com/theforeman/foreman_openscap/pull/505

ares:
needs rebase now :-(

ares:
one more rebase needed

ares:
Weird, it seems that even after filling everything, the submit button does not get enabled in the form. Other form seems to work fine.

xprazak2:
That is really strange, submit button is enabled for me when required fields are filled:

valid-form

invalid-form

ares (approved):
On another setup this works fine, not sure what caused it originally. I'm fine merging this as is, but it would be great to limit the form width the same way the OVAL content form is. Currently it takes 100% of the page which looks weird.

PR #482, opened 2021-06-03
Fixes #32714 - Delete OVAL policies from UI

xprazak2:

PR #485, opened 2021-06-17
Fixes #32823- Delete OVAL Contents from UI

xprazak2:
Reuses logic from https://github.com/theforeman/foreman_openscap/pull/482, so that one should go in first.


webpack/components/withDeleteModal.js
```diff
@@ -0,0 +1,51 @@
+import React, { useState } from 'react';
+import PropTypes from 'prop-types';
+import { translate as __, sprintf } from 'foremanReact/common/I18n';
+import ConfirmModal from './ConfirmModal';
+
+const withDelete = Component => {
```
tbrisker, line 6: would this also be solved by theforeman/foreman#8619?

webpack/helpers/mutationHelper.js
```diff
@@ -0,0 +1,68 @@
...
+    ),
+    onError: onError(showToast, resourceName),
+  };
+
+  return useMutation(mutation, options);
+};
+
+export const submitDelete = (mutation, id) => {
+  mutation({ variables: { id } });
+};
```
tbrisker, line 68: would it make sense to have this in foreman core instead of every plugin using graphql creating their own way?

webpack/helpers/toastHelper.js
```diff
@@ -0,0 +1,3 @@
+import { addToast } from 'foremanReact/redux/actions/toasts';
+
+export const showToast = dispatch => toast => dispatch(addToast(toast));
```
tbrisker, line 3: why not dispatching `addToast` directly?

xprazak2: It would require passing `dispatch` around so I could do `dispatch(addToast({ ... }))`, it just seemed more readable to prepare that in one place and then just call `showToast({ ... })`.

webpack/helpers/mutationHelper.js
xprazak2: Yes, it would definitely make sense. I'll be on a lookout for a feature in core that can use this.

webpack/components/withDeleteModal.js
xprazak2: I hope so.

webpack/components/IndexLayout.js
```diff
@@ -16,6 +17,7 @@ const IndexLayout = ({ pageTitle, children }) => (
     <Helmet>
       <title>{pageTitle}</title>
     </Helmet>
+    <ToastsList />
```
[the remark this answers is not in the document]
xprazak2: Removed

package.json
```diff
@@ -42,6 +42,7 @@
     "jest-svg-transformer": "^1.0.0",
     "jest-transform-graphql": "^2.1.0",
     "prettier": "^1.13.5",
+    "redux-mock-store": "^1.5.4",
```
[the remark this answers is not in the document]
xprazak2: Yes, we should add it.

xprazak2:
Changed to wip so this is be merged only after policy delete.

ares:
needs a rebase now

PR #486, opened 2021-07-01
Fixes #32930 - Add new OVAL content page

xprazak2:
Needs https://github.com/theforeman/foreman/pull/8628 so that permissions can be properly handled, otherwise works.


xprazak2:
Permissions are addressed in https://github.com/theforeman/foreman_openscap/pull/495

ares:
I was able to find this page by visiting `/experimental/compliance/oval_contents/new`, I'd expect the New button on `/compliance/oval_contents` even when there's no other OVAL content. The welcome page doesn't show it though. It's there only if I create at least one content. Also could you please align the radio button with the labels?

alignment

When I'm trying to add it as URL, I'm always told it was successful, however it does not create anything and redirects me to the list where the new content is not created. Also URL would benefit from a hint that shows example, ideally based on https://www.redhat.com/security/data/oval/

Last nitpick is the spacing between the header and the Create button, it would deserve some extra space in between.

Other than that, looks as a solid base.

ares:
> I'm always told it was successful, however it does not create anything

actually it does create the content but it's not shown until I refresh the page

xprazak2:
Changes:

* create button shows on welcome page
* Radio labels are aligned
* OVAL content shows after creation without the need to refresh the page
* placeholder for the URL field
* more space between header, button and table

Before: 
title-before

After:
title-after

ares:
needs a rebase now, sorry :-( ping me right after you do the rebases so it does not wait too long again

PR #488, opened 2021-07-12
Fixes #32988 - Fix apipie doc in oval policies controller

xprazak2:


ares (approved):
Thanks @xprazak2, merging!

PR #489, opened 2021-07-14
Fixes #33015 - Add OVAL Content show page

xprazak2:


ares:
Works well, do you plan to add anything else here? E.g. the delete button, list of policies that use it or is that planned in separate PR?

ares:
Few other random thougst: 
* perhaps also link to "define a new policy based on this content" would be nice :-)
* the URL (for url based content) could be clickable and open a new tab
* the uploaded content should be download-able like we allow for SCAP content, for investigation purposes

xprazak2:
No immediate plans to add more right now, though all the things you mentioned would be nice additions - I'll turn them into Redmine issues for now. I am just not sure about redirecting to the url since those files are bzipped...

ares (approved):
Thanks @xprazak2, merging!

PR #490, opened 2021-07-15
Fixes #33023 - destroy dependent asset_policies

ezr-ondrej:
We are keeping the asset_policies references in database, which prevents policies from ever being deleted.


ares:
Any thoughts on how to cleanup the mess created until this patch?

xprazak2:
It does not work quite as expected, the host now fails to be deleted.

PR #492, opened 2021-07-16
[WIP] Fixes #33041 - Sync OVAL content from policy page

xprazak2:
Needs cleanup and tests,  opening for visibility so that we can decide if we want to go this way.
CPT2107161406-1898x806


ares:
needs a rebase now :-(

ares:
One thing that seems weird is that we update the content on the policy side. Updating the content will have an impact on all policies using the same content, correct? What was the thoughts behind putting it to the policy page? I'm not strictly against, just curious. Also it seems that if the content URL is not valid and the sync therefore fails, it's still showed as successful sync or am I doing something wrong?

xprazak2:
> Updating the content will have an impact on all policies using the same content, correct?
Correct. I think we wanted to emphasize that the policy will be updated with the new content version, but I am not entirely sure about the reasoning that got us there. I can move the sync feature to the content detail if you prefer.

> Also it seems that if the content URL is not valid and the sync therefore fails, it's still showed as successful sync or am I doing something wrong?

It looks like a bug, the date of last sync attempt is shown regardless of the result.

PR #494, opened 2021-07-26
Fixes #33137 - Align with puppet moving to a plugin

xprazak2:
Since the plan is to have foreman_puppet installed for both new and existing users on 3.0, we do not need to switch policy deploy type just yet.

### DIFF
--- a/app/services/foreman_openscap/client_config/puppet.rb
+++ b/app/services/foreman_openscap/client_config/puppet.rb
@@ -10,7 +10,7 @@ module ForemanOpenscap
       end
 
       def available?
-        defined?(Puppetclass)
+        defined?(ForemanPuppet)
       end
 
       def inline_help
@@ -21,13 +21,17 @@ module ForemanOpenscap
         }
       end
 
+      def collection_method
+        :puppetclasses
+      end
+
       def constants
         OpenStruct.new(
           :server_param => 'server',
           :port_param => 'port',
           :policies_param => 'policies',
           :puppetclass_name => 'foreman_scap_client',
-          :config_item_class_name => 'Puppetclass',
+          :config_item_class_name => 'ForemanPuppet::Puppetclass',
           :override_method_name => 'class_params',
           :msg_name => _('Puppet class'),
           :lookup_key_plural_name => _('Smart Class Parameters'),
--- a/.github/workflows/ruby_tests.yml
+++ b/.github/workflows/ruby_tests.yml
@@ -20,6 +20,7 @@ jobs:
         foreman-core-branch: [develop]
         foreman-ansible-branch: [master]
         foreman-rex-branch: [master]
+        foreman-puppet-branch: [master]
         ruby-version: [2.5, 2.6]
     steps:
       - name: Install build packages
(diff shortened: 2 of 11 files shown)


ekohl:
> Since the plan is to have foreman_puppet installed for both new and existing users on 3.0, we do not need to switch policy deploy type just yet.

Note that we do plan on making it possible, but it'll be experimental.

ezr-ondrej (approved):
Looks good to me :+1:

xprazak2:
Test are quite red, though passing for me locally. I'll need to take a closer look.

app/services/foreman_openscap/client_config/puppet.rb
```diff
@@ -21,13 +21,17 @@ def inline_help
         }
       end
 
+      def collection_method
+        :puppetclasses
```
ezr-ondrej, line 25: This will not work, HG, nor Host exposes puppetclasses and all_puppetclasses.
It has moved to `(host || hg).puppet` (what is a Facet for puppet)

ezr-ondrej (changes requested):

app/services/foreman_openscap/client_config/puppet.rb
ezr-ondrej: couldn't we let the config to extract the collection? it would give us more flexibility and we could call the object.puppet.collection :thinking:

xprazak2: We could do that as an emergency solution, it should work with https://github.com/theforeman/foreman_puppet/pull/180

ezr-ondrej:
This will work once the 1.0 puppet is released

lzap:
Tests are failing, any plans to merge this?

xprazak2:
Tests on Jenkins will not pass due to missing dependencies, but GH actions are green, so I think this is ready.

ezr-ondrej (approved):
Looks ready enough to me :+1:

lzap:
Would be fantastic to have it, I am working on some big refactoring of the plugin and green tests would be very welcome :-)

xprazak2:
Thanks everyone!

lzap:
Running tests locally, crossing my fingers! No luck...

```
Traceback (most recent call last):
	6: from /home/lzap/.rbenv/versions/2.7.2/lib/ruby/gems/2.7.0/gems/rake-13.0.3/lib/rake/rake_test_loader.rb:5:in `<main>'
	5: from /home/lzap/.rbenv/versions/2.7.2/lib/ruby/gems/2.7.0/gems/rake-13.0.3/lib/rake/rake_test_loader.rb:5:in `select'
	4: from /home/lzap/.rbenv/versions/2.7.2/lib/ruby/gems/2.7.0/gems/rake-13.0.3/lib/rake/rake_test_loader.rb:17:in `block in <main>'
	3: from /home/lzap/.rbenv/versions/2.7.2/lib/ruby/gems/2.7.0/gems/rake-13.0.3/lib/rake/rake_test_loader.rb:17:in `require'
	2: from /home/lzap/work/foreman_openscap/test/functional/api/v2/compliance/arf_reports_controller_test.rb:1:in `<top (required)>'
	1: from /home/lzap/work/foreman_openscap/test/functional/api/v2/compliance/arf_reports_controller_test.rb:1:in `require'
/home/lzap/work/foreman_openscap/test/test_plugin_helper.rb:8:in `<top (required)>': uninitialized constant ForemanPuppet (NameError)
rake aborted!
```

This fails here:

```
  # Add factories from foreman_ansible
  FactoryBot.definition_file_paths << File.join(ForemanAnsible::Engine.root, '/test/factories')
  FactoryBot.definition_file_paths << File.join(ForemanPuppet::Engine.root, '/test/factories') # HERE
  FactoryBot.reload
```

I think we need a `defined?` there? I also see few tests failing, they are all assuming that Puppet plugin is present. I don't have it installed at the moment.

lzap:
Filed a PR to fix this: https://github.com/theforeman/foreman_openscap/pull/500

PR #493, opened 2021-07-23
Fixes #33126 - Show proper 'permission denied' page

xprazak2:
Needs https://github.com/theforeman/foreman/pull/8692

PR #495, opened 2021-07-30
Refs #32930 - Add permissions for new OVAL content page

xprazak2:
Many changes here are also contained in https://github.com/theforeman/foreman_openscap/pull/493, so that one should probably go in first.


xprazak2:
No longer blocked, should be ready.

ares:
So I have a user with Manager role which I double checked has the `create_oval_contents` permission. However when that user is impersonated, the menu item is not visible at all (perhaps it's only showed to admins) and when I try to enter the page manually at `/experimental/compliance/oval_contents` I get to the index page fine. Clicking the create oval content button then results in Permission denied, but it should work.

Under the user with the viewer role, when I enter the index page, the button is not displayed (correct) and the new form is also not accessible.

So something is wrong for the user with the permission, but then good to be merged.

PR #496, opened 2021-07-30
Fixes #33194 - Show more columns for OVAL content

xprazak2:


ares (approved):
Looks good, merging. In future we may consider combining the two into a single field as it's clear whether it's URL or a file. We could call it source and put the icons if necessary to distinguish. For now it's useful as it is :-)

PR #497, opened 2021-08-02
Fixes #28609 - Remove fix for skipping lone taxonomies assignment

maccelf:
Assigning lone taxonomies was fixed in core, we can now remove the fix we have in arf_reports_controller.

### DIFF
--- a/app/controllers/api/v2/compliance/arf_reports_controller.rb
+++ b/app/controllers/api/v2/compliance/arf_reports_controller.rb
@@ -140,12 +140,6 @@ module Api
             super
           end
         end
-
-        protected
-
-        def assign_lone_taxonomies
-          # do not assign lone taxonomies to arf report
-        end
       end
     end
   end


ares:
I think that makes sense, the original fix for the record is https://github.com/theforeman/foreman/pull/7275/files which now shortcircuit the `assign_lone_taxonomies` for reports from core if I understand correctly.

[ok to test]

@maccelf could you please confirm that's the logic behind this change?

maccelf:
> I think that makes sense, the original fix for the record is https://github.com/theforeman/foreman/pull/7275/files which now shortcircuit the `assign_lone_taxonomies` for reports from core if I understand correctly.
> 
> [ok to test]
> 
> @maccelf could you please confirm that's the logic behind this change?

@ares yes, i think is that the logic behind this change

xprazak2:
Looks good, the tests will not pass for now due to puppet changes.

ares:
Ok, :+1: from me, leaving up to @xprazak2 to merge when we're ready.

xprazak2:
Thanks!

PR #500, opened 2021-08-23
Refs #33137 - tests work without Puppet plugin

lzap:
Tests do fail without Puppet plugin installed. This fixes it.


xprazak2:
Thanks!

PR #503, opened 2021-08-27
Fixes #33362 - Edit OVAL policy attributes

xprazak2:


ares:
It would be good to add some spacing between the tabs and the details, the Name is too close to it IMHO
oval_edit

also one unfortunate thing is the actual editting, when I hit Enter, it seems to submit the form the old way. Click the edit icon, perform some change and hit Enter. The page will reload and the modification is not performed.

And cancelling does not work well, try opening the edit form, then delete the content, click the checkmark to save it, you see the validation error. Now click cancel (the cross) and from now on, you see "None provided" as the value. Reloading the page of course load the right value again.

ares:
I just realized the first comment is unrelated to the editing as it's the same issue even without edit links. It would be good to track that though and fix it, even if it's out of this PR. It seems the same thing is present also for contents.

PR #508, opened 2021-10-04
Fixes #33624 - Show list of CVEs on Host detail page

xprazak2:


ares:
[test]

ares:
[test foreman_openscap]

app/models/foreman_openscap/cve.rb
```diff
@@ -5,6 +5,7 @@ class Cve < ApplicationRecord
     has_many :oval_policies, :through => :host_cves
 
     scoped_search :relation => :host_cves, :on => :oval_policy_id, :rename => :oval_policy_id, :complete_value => false
+    scoped_search :relation => :host_cves, :on => :host_id, :rename => :host_id, :complete_value => false
```
ares, line 8: `:only_explicit => true` would be useful to avoid searching in this value without specifying attribute name, if you fix that, please also add the same to the one above

ares (commented):
tests seems to be failing, one comment re scoped search definition

webpack/components/OpenscapHostDetail/index.js
```diff
@@ -0,0 +1,94 @@
...
+
+  if (status === STATUS.ERROR) {
+    return (
+      <EmptyState
+        error
+        title={__('Error!')}
+        body={__('There has been an error when loading host information.')}
+      />
+    );
+  }
```
amirfefer, line 41: can we use `SkeletonLoader` instead?

webpack/components/OpenscapHostDetail/index.js
```diff
@@ -0,0 +1,94 @@
...
+              id={response.id}
+              history={history}
+              router={router}
+              match={match}
+              searchKey="host_id"
+              linkCell={hostPageLinkCell}
+            />
+          </TabLayout>
+        </Route>
+      </Switch>
```
amirfefer, line 71: how about extracting these routes to a separate component i.e `SecondaryTabRoutes`?

webpack/components/OpenscapHostDetail/index.js
```diff
@@ -0,0 +1,94 @@
...
+      />
+    );
+  }
+
+  const cvesPath = 'cves';
+
+  return (
+    <React.Fragment>
+      <Tabs activeKey={setActiveKey(location)} isSecondary>
+        <Tab
```
amirfefer, line 48: is there a plan for adding more tabs?

webpack/helpers/tableHelper.js
```diff
@@ -7,3 +7,11 @@ export const linkCell = (path, text) => (
     <Link to={path}>{text}</Link>
   </TableText>
 );
+
+export const hostPageLinkCell = (path, text, history) => (
```
amirfefer, line 11: After some thinking, we don't need the `router` prop from the slot.
foreman's client routing uses `connected-react-router` so we can dispatch an action for triggering a url change.
we have already a helper for that under `foreman_navigation` (pushUrl)

amirfefer: agree, the `pushUrl` helper  should be imported directly from `foremanReact`

amirfefer: I mean it should be :) 
meanwhile, we can consume `push` directly from `connected-react-router` in plugins.

xprazak2: I tried that, the following does not work as expected:

```js
import { push } from 'connected-react-router'
//
<a href="#" onClick={() => push(path)}>
  {text}
</a>
```

TheRedGreek:
Is there any update on this? Can we merge it, if you need help testing, maybe I can contribute?

PR #509, opened 2021-10-06
Fix tests

xprazak2:
Make sure the :lab_features setting is correctly found
or default to false so that db constraint is not violated.


xprazak2:
Test on Jenkins have had a broken setup for a while now - the error is caused because smart_proxy_dynflow gem is pulled into the bundle via *core gems. The GH action tests pass though.

PR #512, opened 2021-11-10
Fixes #33755 - Do not show errors on inline edit reopen

xprazak2:


ares (approved):
Fixes the issue, thanks! Merging now.

PR #520, opened 2021-12-02
Fixes #34069 - Stop using deprecated redux toast actions

ofedoren:


tbrisker (changes requested):
I think you forgot to commit the new file, but I would suggest getting rid of the helper and just dispatching the action directly. It would also allow getting rid of the wrapped component.

ofedoren:
> I think you forgot to commit the new file, but I would suggest getting rid of the helper and just dispatching the action directly. It would also allow getting rid of the wrapped component.

I didn't forget, this repo has two alsmost the same helper files: https://github.com/theforeman/foreman_openscap/blob/master/webpack/helpers/toastHelper.js
https://github.com/theforeman/foreman_openscap/blob/master/webpack/helpers/toastsHelper.js

Okay then, I'll try to get rid of both of them :)

tbrisker (approved):
never mind, looks like we had a duplicate helper :\ I still think we should drop it and use the action directly, but doesn't have to be in this pr

PR #522, opened 2022-01-11
rubocop fixes

Ron-Lavi:
the error was `Rails/HasManyOrHasOneDependent: Specify a :dependent option.`


.rubocop_todo.yml
```diff
@@ -178,3 +178,6 @@ Rails/ReadWriteAttribute:
 Rails/TimeZone:
   Exclude:
     - 'app/models/foreman_openscap/arf_report.rb'
+
+Rails/HasManyOrHasOneDependent:
+  Enabled: false
```
Ron-Lavi, line 183: hey @ares should we use `Exclude` here instead?

PR #524, opened 2022-02-03
Fixes #34385 - more reliable way of detecting Ansible plugin

ares:
After the ForemanAnsible namespace has been defined in the Foreman core,
we need to use something more reliable to detect the plugin presence.
The version is always defined only in the plugin gem.

### DIFF
--- a/app/services/foreman_openscap/client_config/ansible.rb
+++ b/app/services/foreman_openscap/client_config/ansible.rb
@@ -16,7 +16,7 @@ module ForemanOpenscap
       end
 
       def available?
-        defined?(ForemanAnsible)
+        Foreman::Plugin.installed?("foreman_ansible")
       end
 
       def inline_help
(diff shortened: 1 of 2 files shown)


app/services/foreman_openscap/client_config/ansible.rb
```diff
@@ -16,7 +16,7 @@ def type
       end
 
       def available?
-        defined?(ForemanAnsible)
+        defined?(ForemanAnsible::VERSION)
```
evgeni, line 19: Shouldn't this use `Foreman::Plugin.installed?("foreman_ansible")` instead? I mean, we have a Plugin Registry for a reason? ;)

Ron-Lavi:
@evgeni PR using ruby 2.7 is merged,
can you also rebase?

app/services/foreman_openscap/client_config/ansible.rb
ares: I had the impression the plugin may be registered later, when it's engine is processed. And I am also under the impression migrations don't have the whole environment loaded at this point. But let's try and see if that works in migrations.

ares:
Rebased, let's see the tests results

Ron-Lavi (approved):
Thanks @ares @evgeni ! LGTM

app/services/foreman_openscap/client_config/ansible.rb
ezr-ondrej: I'm very happy we enden up using our own framework for checking wether our plugin is installed ^_^

PR #525, opened 2022-02-03
use Ruby 2.7 for tests

evgeni:
Foreman is currently using 2.5 and 2.7 in prod deployments, not 2.6


Ron-Lavi (approved):
Thanks @evgeni !

PR #527, opened 2022-02-06
Fixes #34274 - consume PF4 pagination from core

Ron-Lavi:


amirfefer (approved):
Tested, it works well and tests aren't broken 👍 
Though, we don't really need these props
```js
   onSetPage={handlePageSelected}
   onPerPageSelect={handlePerPageSelected}
```
It works well without these, but it seems that the tests use it for mocking the `first/last` page.

ares:
Thanks @Ron-Lavi and @amirfefer, megring!

PR #529, opened 2022-11-24
i18n - pulling from tx

ekohl:


ekohl:
Note this doesn't include `ka` for which there is a translation.

PR #530, opened 2022-12-02
CI fixes

ekohl:


ekohl:
I'm not sure why this still fails on Jenkins, but it's an improvement so I'm merging this.

PR #546, opened 2023-09-27
Fixes #36738 - Add remediation wizard

ofedoren:
There are still some TODOs (marked in the code):
 - [ ] Show the wizard for failed rules only.
 - [ ] Fix pagination in review hosts step. I'm not sure why it's not rendered properly, maybe @MariaAga could help?..
 - [ ] What permissions should be checked and where?
 - [ ] Maybe we could use `Log`/`Message` model directly to load data faster instead of loading whole report.
 - [ ] `ansible_run_openscap_remediation` feature template should probably live in foreman_ansible repo.
 
 To test this, you should have a host with policy associated and an arf report for that host, navigate to the `/compliance/arf_reports/<id>`, there is a new `Remediate` entry in action buttons for every rule.


app/views/job_templates/run_openscap_remediation_-_script_default.erb
```diff
@@ -0,0 +1,24 @@
...
+  required: true
+- name: reboot
+  description: Indicate wether the host should be rebooted after the remediation
+  input_type: user
+  required: false
+%>
+<%= input("command") %>
+<% if truthy?(input('reboot')) -%>
+echo "A reboot is required to finish the remediation. The system is going to reboot now."
+reboot
```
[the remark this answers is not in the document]
ofedoren: That makes much more sense to just schedule the reboot instead of doing it instantly... Thanks :)

webpack/components/OpenscapRemediationWizard/steps/ReviewHosts.js
```diff
@@ -0,0 +1,199 @@
...
+              <ToolbarItem>
+                <Spinner size="sm" />
+              </ToolbarItem>
+            )}
+          </ToolbarGroup>
+        </ToolbarContent>
+      </Toolbar>
+      <Table
+        isEmbedded
+        isActionable={false}
```
MariaAga, line 175: isActionable is not a prop, fixed the empty actions here: https://github.com/theforeman/foreman/pull/10082

MariaAga (changes requested):

ofedoren:
Thanks a lot, @MariaAga, should be fixed now.

webpack/components/OpenscapRemediationWizard/OpenscapRemediationSelectors.js
```diff
@@ -0,0 +1,18 @@
...
+  selectAPIResponse,
+  selectAPIStatus,
+} from 'foremanReact/redux/API/APISelectors';
+import { STATUS } from 'foremanReact/constants';
+import {
+  REPORT_LOG_REQUEST_KEY,
+} from './constants';
+
+export const selectLogResponse = state =>
+  selectAPIResponse(state, REPORT_LOG_REQUEST_KEY) || {};
```
MariaAga, line 12: this condition is already done in selectAPIResponse

webpack/components/OpenscapRemediationWizard/OpenscapRemediationSelectors.js
```diff
@@ -0,0 +1,18 @@
+import {
+  selectAPIError,
+  selectAPIResponse,
+  selectAPIStatus,
+} from 'foremanReact/redux/API/APISelectors';
+import { STATUS } from 'foremanReact/constants';
+import {
+  REPORT_LOG_REQUEST_KEY,
```
MariaAga, line 8: spacing for eslint

webpack/components/OpenscapRemediationWizard/steps/ReviewHosts.js
```diff
@@ -0,0 +1,198 @@
...
+  const defaultParams = {
+    search: defaultSearch,
+  };
+
+  const [params, setParams] = useState(defaultParams);
+
+  const searchProps = getControllerSearchProps('hosts');
+  searchProps.autocomplete.searchQuery = defaultSearch;
+  searchProps.autocomplete.url = '../../hosts/auto_complete_search'; // TODO: find a way to not use relative path
+  searchProps.disabled = true; // This is to force hosts to be searched by currently remediated rule only
```
MariaAga, line 45: If the search is always disabled, why is there an "onSearch" function, or a search bar in general?

webpack/components/OpenscapRemediationWizard/steps/ReviewHosts.js
```diff
@@ -0,0 +1,198 @@
...
+
+  useEffect(() => {
+    if (selectedCount) setHostIdsParam(fetchBulkParams());
+  }, [selectedCount, fetchBulkParams, setHostIdsParam]);
+
+  const selectionToolbar = (
+    <ToolbarItem key="selectAll">
+      <SelectAllCheckbox
+        {...{
+          selectAll: noop, // I don't think it really can select all since ids from other pages are still need to be loaded/fetched
```
MariaAga, line 106: We can send a search query to the job invocation maybe?

MariaAga: or just have selectAll: selectPage instead, as now the button doesn't select anything and its confusing

webpack/components/OpenscapRemediationWizard/steps/ReviewHosts.js
```diff
@@ -0,0 +1,198 @@
...
+          selectPage,
+          selectNone: noop, // we shouldn't allow to select none since it migth lead to selecting a wrong host
```
MariaAga, line 108: Select none should select the host the button was clicked on, of have some kind of action

webpack/components/OpenscapRemediationWizard/steps/ReviewHosts.js
ofedoren: Initially I wanted to allow the search, but after more thinking I've decided to just disable it and show users why these hosts are shown exactly. Probably can be done via simple note entry.

webpack/components/OpenscapRemediationWizard/steps/ReviewHosts.js
ofedoren: Hmm... Select All option should not be available if `noop` is passed: https://github.com/theforeman/foreman/blob/181e5040e6026caa74daa2f30809ceba7dc80bf0/webpack/assets/javascripts/react_app/components/PF4/TableIndexPage/Table/SelectAllCheckbox/index.js#L27

webpack/components/OpenscapRemediationWizard/steps/ReviewHosts.js
ofedoren: I've tried that with calling imported selectNone and then using selectOne as part of selectNone, but for some reason it didn't want to select the host, so I've decided to hide Select None option as with Select All. That'd need a patch in Foreman though...

webpack/components/OpenscapRemediationWizard/steps/ReviewRemediation.js
```diff
@@ -0,0 +1,179 @@
...
+  const [copied, setCopied] = useState(false);
+  const selectedFix = findFixBySnippet(fixes, snippet);
+  const snippetText = selectedFix?.full_text;
+  // can be one of null, "true", "false"
+  // if null, it may indicate that reboot might be needed
+  const checkForReboot = () => !some(fixes, { reboot: 'false' });
+  const isRebootRequired = () => some(fixes, { reboot: 'true' });
+
+  const copyToClipboard = (e, text) => {
+    navigator.clipboard.writeText(text.toString());
```
MariaAga, line 48: (Also relevant to all other copyToClipboards we have), we can assume were always in a in secure context (HTTPS), right?  https://developer.mozilla.org/en-US/docs/Web/API/Navigator/clipboard

ofedoren: Well, we should. At least with self-signed certificates.

MariaAga: Great!

webpack/components/OpenscapRemediationWizard/steps/ReviewHosts.js
MariaAga: I tried:
``` js
selectNone: () => {
            selectNone();
            selectOne(true, hostId);
          },
```
and it worked fine, still its a bit confusing that the main checkbox doesn't select the whole page, but I'm sure if its a fix for the plugin or foreman core

ofedoren: That's just weird... Probably the new way of testing UI is through incognito mode :/ Never had issues with something being cached before. Or maybe that was something else. I've just checked again and it seems to be working indeed.

MariaAga: Jeremey said that he had this issue, and he checked the disable cache box in the dev tools, network tab and it got resolved, maybe it will help? (I have it on for years so I never notice these issues ): )

ofedoren:
@MariaAga, I've returned `Select All` option, so it should be now less weird to use the select box.

ofedoren:
@lhellebr, I think you might want to check the current state, rpm build for rhel8 is ready: https://dashboard.packit.dev/results/copr-builds/1425886

MariaAga:
Thanks!
all selected is lost when moving between steps (select all in review hosts, click next, click back, only 1 selected)

lhellebr:
I have tested this PR together with https://github.com/theforeman/foreman/pull/10113 and without any deeper testing, I can confirm the workflow works.

adamruzicka (approved):
The only test failures seem to be coming from oval, which is not a thing that should be solved here

MariaAga (approved):

webpack/components/OpenscapRemediationWizard/steps/ReviewRemediation.js
```diff
@@ -0,0 +1,186 @@
...
+
+  const onCopyClick = (e, text) => {
+    copyToClipboard(e, text);
+    setCopied(true);
+  };
+
+  const description =
+    method === 'manual'
+      ? __(
+          'Please review the remediation snippet and apply to the host manually.'
```
asteflova, line 62:
```suggestion
          'Review the remediation snippet and apply it to the host manually.'
```

webpack/components/OpenscapRemediationWizard/steps/ReviewRemediation.js
```diff
@@ -0,0 +1,186 @@
...
+        )
+      : __(
+          'Please review the remediation snippet that will be applied to selected host(s).'
```
asteflova, line 65:
```suggestion
          'Review the remediation snippet that will be applied to selected host(s).'
```

webpack/components/OpenscapRemediationWizard/steps/ReviewRemediation.js
```diff
@@ -0,0 +1,186 @@
...
+            <GridItem span={12} rowSpan={1}>
+              <Alert
+                ouiaId="reboot-alert"
+                variant={isRebootRequired() ? 'warning' : 'info'}
+                title={rebootAlertTitle}
+              >
+                {method === 'manual' ? null : (
+                  <p>
+                    {__(
+                      'You can tick the checkbox below to reboot the system(s) automatically after the remediation is applied.'
```
asteflova, line 155: You probably don't need this. The checkbox that says "Reboot the system(s)" is self-explanatory.

Removing this line from the info box might actually help because this part of the wizard is a bit too complex already; simplifying it by dropping the line would help.

webpack/components/OpenscapRemediationWizard/steps/ReviewRemediation.js
```diff
@@ -0,0 +1,186 @@
...
+    method === 'manual'
+      ? __(
+          'Please review the remediation snippet and apply to the host manually.'
+        )
+      : __(
+          'Please review the remediation snippet that will be applied to selected host(s).'
+        );
+
+  const rebootAlertTitle = isRebootRequired()
+    ? __('A reboot is required.')
```
asteflova, line 69:
```suggestion
    ? __('A reboot is required after applying remediation .')
```

webpack/components/OpenscapRemediationWizard/steps/ReviewRemediation.js
```diff
@@ -0,0 +1,186 @@
...
+    : __('A reboot might be needed.');
```
asteflova, line 70:
```suggestion
    : __('A reboot might be required after applying remediation.');
```

asteflova (commented):
I found a couple of places where it would be nice to adjust the wording so that we don't have to explain in more detail in documentation. And I'm adding a few other very minor suggestions as well.

Feel free to reject any or all of my suggestions (but if possible, please explain why you're rejecting them so that I can learn :))

Overall, I like how you're doing this because in the docs, I really will be able to just explain where to find the wizard and the web UI will guide users through the whole process without requiring additional documentation. Great!

adamruzicka:
Thank you @ofedoren , @MariaAga, @lhellebr & @asteflova !

PR #547, opened 2023-10-17
Fixes #36838 - Make doc links point to the new docs

adamruzicka:
I know this isn't exactly clean, but this should get cherrypicked all the way back into a 3.6-compatible branch so we can't really add some nice infrastructure into Foreman itself and then use it here without making things hard for ourselves.

### DIFF
--- a/app/helpers/foreman_openscap_helper.rb
+++ b/app/helpers/foreman_openscap_helper.rb
@@ -16,7 +16,7 @@ module ForemanOpenscapHelper
   end
 
   def doc_flavor
-    ::Foreman::Plugin.installed?('katello') ? 'katello' : 'foreman'
+    ForemanOpenscap.with_katello? ? 'katello' : 'foreman'
   end
 
   def scap_root_url
--- a/app/helpers/foreman_openscap_helper.rb
+++ b/app/helpers/foreman_openscap_helper.rb
@@ -1,12 +1,8 @@
 require 'foreman_openscap/version'
 
 module ForemanOpenscapHelper
-  def scap_doc_link(section = '', text = _('documentation'))
-    link_to(
-      text,
-      scap_doc_url(section),
-      :rel => 'external noopener noreferrer', :target => '_blank'
-    )
+  def scap_doc_button(section)
+    documentation_button(section, root_url: scap_doc_url)
   end
 
   def scap_doc_url(section = '')
@@ -14,6 +10,8 @@ module ForemanOpenscapHelper
 
     documentation_url(section, root_url: scap_root_url)
   end
+
+  private
 
   def doc_flavor
     ForemanOpenscap.with_katello? ? 'katello' : 'foreman-el'
--- a/app/helpers/foreman_openscap_helper.rb
+++ b/app/helpers/foreman_openscap_helper.rb
@@ -15,11 +15,15 @@ module ForemanOpenscapHelper
     documentation_url(section, root_url: scap_root_url)
   end
 
+  def doc_flavor
+    ::Foreman::Plugin.installed?('katello') ? 'katello' : 'foreman'
+  end
+
   def scap_root_url
     @scap_root_url ||= begin
       version = SETTINGS[:version]
       version = version.tag == 'develop' ? 'nightly' : version.short
-      "https://docs.theforeman.org/#{version}/Managing_Security_Compliance/index-foreman-el.html#"
+      "https://docs.theforeman.org/#{version}/Managing_Security_Compliance/index-#{doc_flavor}-el.html#"
     end
   end
 end
--- a/app/helpers/foreman_openscap_helper.rb
+++ b/app/helpers/foreman_openscap_helper.rb
@@ -16,14 +16,14 @@ module ForemanOpenscapHelper
   end
 
   def doc_flavor
-    ForemanOpenscap.with_katello? ? 'katello' : 'foreman'
+    ForemanOpenscap.with_katello? ? 'katello' : 'foreman-el'
   end
 
   def scap_root_url
     @scap_root_url ||= begin
       version = SETTINGS[:version]
       version = version.tag == 'develop' ? 'nightly' : version.short
-      "https://docs.theforeman.org/#{version}/Managing_Security_Compliance/index-#{doc_flavor}-el.html#"
+      "https://docs.theforeman.org/#{version}/Managing_Security_Compliance/index-#{doc_flavor}.html#"
     end
   end
 end
--- a/app/helpers/foreman_openscap_helper.rb
+++ b/app/helpers/foreman_openscap_helper.rb
@@ -18,7 +18,7 @@ module ForemanOpenscapHelper
   def scap_root_url
     @scap_root_url ||= begin
       version = ForemanOpenscap::VERSION.split('.')[0..-2].join('.')
-      "https://theforeman.org/plugins/foreman_openscap/#{version}/index.html"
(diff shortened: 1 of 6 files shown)


app/helpers/foreman_openscap_helper.rb
```diff
@@ -15,10 +15,15 @@ def scap_doc_url(section = '')
     documentation_url(section, root_url: scap_root_url)
   end
 
+  def doc_flavor
+    ::Foreman::Plugin.installed?('katello') ? 'katello' : 'foreman'
```
ofedoren, line 19: What if we add check-for-katello-plugin helper? Something like 
https://github.com/theforeman/foreman_openscap/blob/master/lib/foreman_openscap/engine.rb#L303-L305

adamruzicka: Huh, we already have that https://github.com/theforeman/foreman_openscap/blob/master/lib/foreman_openscap/engine.rb#L2 ?

app/helpers/foreman_openscap_helper.rb
```diff
@@ -15,10 +15,15 @@ def scap_doc_url(section = '')
...
+    ForemanOpenscap.with_katello? ? 'katello' : 'foreman'
+  end
+
   def scap_root_url
     @scap_root_url ||= begin
-      version = ForemanOpenscap::VERSION.split('.')[0..-2].join('.')
-      "https://theforeman.org/plugins/foreman_openscap/#{version}/index.html"
+      version = SETTINGS[:version]
+      version = version.tag == 'develop' ? 'nightly' : version.short
+      "https://docs.theforeman.org/#{version}/Managing_Security_Compliance/index-#{doc_flavor}-el.html#"
```
ofedoren, line 26:
```suggestion
      "https://docs.theforeman.org/#{version}/Managing_Security_Compliance/index-#{doc_flavor}.html#"
```

ofedoren (changes requested):
Thanks, @adamruzicka, LGTM and works, but needs a small fix, see inline. It seems we don't need to explicitly mention `el` for https://docs.theforeman.org/.

app/helpers/foreman_openscap_helper.rb
ofedoren: Not sure I understand...
[Peek 2023-10-19 15-20.webm](https://github.com/theforeman/foreman_openscap/assets/32508194/589b01e8-d03d-4b8d-b4ea-3adefc9bf16f)

adamruzicka: Oh, right. Because katello doesn't run on non-el, but foreman does. Will fix

app/helpers/foreman_openscap_helper.rb
```diff
@@ -15,10 +15,15 @@ def scap_doc_url(section = '')
     documentation_url(section, root_url: scap_root_url)
   end
 
+  def doc_flavor
+    ForemanOpenscap.with_katello? ? 'katello' : 'foreman-el'
```
ekohl, line 19: My biggest concern is that the OpenSCAP guide is not published in stable branches as `foreman-el`:

https://github.com/theforeman/foreman-documentation/blob/5a811af4a30310b96a15e3bec708079e04256c31/guides/doc-Managing_Security_Compliance/master.adoc#L14-L18

This results in linking to https://docs.theforeman.org/3.8/Managing_Security_Compliance/index-foreman-el.html (for Foreman 3.8).

This is because it hasn't been properly reviewed. You can see examples such as https://docs.theforeman.org/nightly/Managing_Security_Compliance/index-foreman-el.html#Deploying_a_Policy_in_a_Host_Group_Using_Ansible_security-compliance

> You have enabled and synced the operating system repositories to Foreman, and enabled them on the hosts:

Content syncing is not available without Foreman, so that needs to be addressed.

We do publish the nightly docs, just so you can at least review some of it, even if it's broken.

ekohl (commented):
I'd imagine this also needs changes in foreman_theme_satellite, but looking at https://github.com/RedHatSatellite/foreman_theme_satellite/blob/develop/lib/foreman_theme_satellite/documentation.rb (so happy that it's public now) I don't see it, so perhaps downstream also linked to unbranded docs.

app/views/arf_reports/welcome.html.erb
```diff
@@ -9,6 +9,6 @@
     <%= _("You don't seem to have any ARF report. ARF report is a summary of a single scan occurrence on a particular host for a given Compliance Policy.") %></br>
   </p>
   <div class="blank-slate-pf-main-action">
-    <%= link_to _('Documentation'), documentation_url("4.4ARFReports", :root_url => "https://www.theforeman.org/plugins/foreman_openscap/0.8/index.html#"), :rel => 'external', :class => 'btn btn-primary btn-lg'  %>
+    <%= link_to _('Documentation'), documentation_url("Monitoring_Compliance_security-compliance", :root_url => scap_doc_url), :rel => 'external', :class => 'btn btn-primary btn-lg'  %>
```
ekohl, line 12: Shouldn't this use `scap_doc_url`?
```suggestion
    <%= link_to _('Documentation'), scap_doc_url("Monitoring_Compliance_security-compliance"), :rel => 'external', :class => 'btn btn-primary btn-lg'  %>
```

For a moment I thought about using `scap_doc_link` but that looks to be not flexible enough.

app/views/policies/index.html.erb
```diff
@@ -3,7 +3,7 @@
 
 <% title_actions(
   display_link_if_authorized(_("New Compliance Policy"), hash_for_new_policy_path, :class => "btn btn-primary"),
-    documentation_button('#4.2Creatingpolicywizard', :root_url => scap_doc_url)
+    documentation_button('Managing_Compliance_Policies_security-compliance', :root_url => scap_doc_url)
```
ekohl, line 6: Would it make sense to create `scap_documentation_button` for now?

ekohl (changes requested):
I think `app/views/policies/steps/_deployment_options_form.html.erb` is still the old link:
```console
$ rg scap_doc_link
app/views/policies/steps/_deployment_options_form.html.erb
7:    <%= _('Please make sure you understand them by reading our') %> <%=scap_doc_link('#2.3Policydeploymentoptions') %>.

app/helpers/foreman_openscap_helper.rb
4:  def scap_doc_link(section = '', text = _('documentation'))
```

I'd consider making `scap_doc_link` more flexible or dropping it.

app/views/policies/steps/_deployment_options_form.html.erb
```diff
@@ -4,7 +4,7 @@
   <div class="alert alert-info" id="scap-deployment-options-info-banner">
     <span class="pficon pficon-info"></span>
     <strong><%= _('There are significant differences in deployment options.') %></strong>
-    <%= _('Please make sure you understand them by reading our') %> <%=scap_doc_link('#2.3Policydeploymentoptions') %>.
+    <%= _('Please make sure you understand them by reading our') %> <%= link_to(_('documentation'), scap_doc_url('deploying-compliance-policies_security-compliance'), :rel => 'external noopener noreferrer', :target => '_blank') %>.
```
ekohl, line 7: Translation wise this isn't great. Other languages may put the word `documentation` at another place, but perhaps we should track it elsewhere.

ekohl (approved):

app/views/policies/steps/_deployment_options_form.html.erb
adamruzicka: https://projects.theforeman.org/issues/36865

ofedoren (approved):
Thanks, @adamruzicka and @ekohl!

Sorry for waiting, LGTM, let's get this in.

PR #548, opened 2023-10-31
Convert to theforeman/actions

ekohl:
Currently a draft since it doesn't call the core extensions yet.

### DIFF
--- a/.github/workflows/ruby_tests.yml
+++ b/.github/workflows/ruby_tests.yml
@@ -1,77 +1,16 @@
+---
 name: Ruby Testing
+
 on:
+  push:
+    branches:
+      - master
+      - '*-stable'
   pull_request:
-env:
-  BUNDLE_WITHOUT: journald:development:console:libvirt
-  RAILS_ENV: test
-  DATABASE_URL: postgresql://postgres:@localhost/test
-  DATABASE_CLEANER_ALLOW_REMOTE_DATABASE_URL: true
+
 jobs:
-  test_ruby:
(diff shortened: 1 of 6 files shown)


ekohl:
I noticed there is a RuboCop config, but nothing calls it today. Or at least, when I activated it, it didn't pass.

.github/workflows/ruby_tests.yml
```diff
@@ -1,79 +1,17 @@
...
-      - name: Run migrations
-        run: |
-          bundle exec rake db:create
-          bundle exec rake db:migrate
-      - name: Run plugin tests
-        run: |
-          bundle exec rake test:foreman_openscap
-      - name: Run selected core tests
-        run: bundle exec rake test:foreman_openscap_extensions
+  # TODO: bundle exec rake test:foreman_openscap_extensions
```
evgeni, line 12: Mhh, `matrix_include` of the action doesn't allow us to extend the `tests` list? Sad.

ofedoren: Could we do something like this https://github.com/theforeman/foreman_openscap/pull/557/files#diff-c105fe0c2711318227c0cf6c915452e4be538d8a7d39a7fa7f387fef15a63311R89? I mean that task is only for these files. Maybe for now we could include them in the main test task?

ekohl: Done so now. I think your had an issue with globbing, not really sure.

ofedoren (approved):
Thanks, @ekohl, @evgeni!

PR #552, opened 2023-11-08
Fixes #36906 - fix bulk Compliance Policy for all hosts

MariaAga:
Updating the function to match
https://github.com/theforeman/foreman/blob/8930302fa46af140b636c6e4776d1caa0e9dbdf9/app/controllers/hosts_controller.rb#L780
I didn't find a way to just use it from the host controller since its private.
The issue for all hosts selected was that the params was `search=` and it just redirected to the hosts page inside the action modal, instead of putting all the hosts into `@hosts`

### DIFF
--- a/app/controllers/policies_controller.rb
+++ b/app/controllers/policies_controller.rb
@@ -117,13 +117,18 @@ class PoliciesController < ApplicationController
     @tailoring_file = ForemanOpenscap::TailoringFile.find(params[:tailoring_file_id]) if params[:tailoring_file_id].present?
   end
 
+  def multiple_with_filter?
+    params.key?(:search)
+  end
+  
   def find_multiple
-    # Lets search by name or id and make sure one of them exists first
-    if params[:host_ids].present?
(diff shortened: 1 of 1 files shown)


ofedoren (commented):
Thanks, @MariaAga, seems to fix... an unexpected behavior?..

Before:
ScreenShot-1699546117641
After:
ScreenShot-1699546611884

In both cases it seemed to be working, but... differently.

I mean, it seems to unassign the policy, but currently it sends the host ids to the server, but with the fix I can't find any host ids in the request logs...

@adamruzicka, could you take a look as well?

adamruzicka:
Seems to work for me. If I select hosts explicitly, `host_id` parameter(s) get sent over. If I select all without searching, there's no parameter at all, if I search and then select all, the search is sent over as `search`. The backend honors these parameters (or lack of thereof) and (un)assigns the policy to the matched hosts

ofedoren (approved):
If so, I don't see anything to be blocked on, let's get this in.

Thanks, @MariaAga and @adamruzicka.

PR #555, opened 2024-01-04
Fixes #37027 - Set up sp-reference to scap proxy

adamruzicka:


ofedoren (approved):
Thanks, @adamruzicka!

PR #556, opened 2024-01-05
Pin bundler to allow tests to run

adamruzicka:


adamruzicka:
Oh, right, this will need to happen in the jenkins repo

PR #557, opened 2024-01-15
Test on Ruby 3

ofedoren:
Please, don't merge just yet. Foreman on Ruby 3.0 is still in progress.


ekohl:
It would be nice if this was based on https://github.com/theforeman/foreman_openscap/pull/548. Given today RuboCop isn't checked, I'd split that off to a separate effort.

ofedoren:
I guess it can be closed, nothing broken, nothing to add :/ access control test could/should be fixed by shared workflow?..

PR #561, opened 2024-03-08
Test only against node 14

adamruzicka:


ofedoren:
Thanks, @adamruzicka, somehow I missed your PR :/ Maybe we should add node 18 since Foreman run on it as well?

ofedoren (approved):
Thanks, @adamruzicka, fails are expected :/

PR #563, opened 2024-03-12
Fixes #33035 - Use foremanUrl for urls

adamruzicka:


ofedoren:
It looks like all of these helper methods are related to OVAL pages. Maybe we could simply remove the whole file as part of OVAL removals?

adamruzicka:
I must admit I didn't look too closely at pruning dead code in the js lands, but yeah, sure.

PR #567, opened 2024-04-30
Fixes #37396 - Fix autocomplete for ArfReports

adamruzicka:
The code expected that a host would have up to one lifecycle environment, which doesn't hold anymore on recent versions of katello. This tripped up scoped_search which couldn't find the right association to search on.

### DIFF
--- a/app/models/concerns/foreman_openscap/compliance_status_scoped_search.rb
+++ b/app/models/concerns/foreman_openscap/compliance_status_scoped_search.rb
@@ -109,7 +109,11 @@ module ForemanOpenscap
 
     included do
       if ForemanOpenscap.with_katello?
-        has_many :lifecycle_environments, :through => :host
+        has_one :content_facet, :through => :host
+        has_many :content_view_environment_content_facets, through: :content_facet, class_name: 'Katello::ContentViewEnvironmentContentFacet'
+        has_many :content_view_environments, through: :content_view_environment_content_facets
+        has_many :content_views, through: :content_view_environments
+        has_many :lifecycle_environments, through: :content_view_environments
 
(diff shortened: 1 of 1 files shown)


pondrejk:
checked on patched stream from the functional side, before:

Screenshot from 2024-04-30 15-00-59

after:

Screenshot from 2024-04-30 15-18-11

pondrejk (approved):

ofedoren:
Thanks, @adamruzicka and @pondrejk !

Although, it fixes the initial issues, I guess we still might have problem that search by lifecycle env won't work:
ScreenShot-1715684340333

This can be also problem on my machine, but if you can reproduce it, do we want to fix that as part of this PR or should we deal with that later?

app/models/concerns/foreman_openscap/compliance_status_scoped_search.rb
```diff
@@ -109,11 +109,15 @@ def apply_condition(scope, negate, conditions)
 
     included do
       if ForemanOpenscap.with_katello?
-        has_one :lifecycle_environment, :through => :host
+        has_one :content_facet, :through => :host
+        has_many :content_view_environment_content_facets, through: :content_facet, class_name: 'Katello::ContentViewEnvironmentContentFacet'
+        has_many :content_view_environments, through: :content_view_environment_content_facets
+        has_many :content_views, through: :content_view_environments
```
ofedoren, line 115: I thought you'll fix that in the way it's done for the other searches through `ext_method`, but that's seems fine as well :)

ofedoren (approved):
Thanks, @adamruzicka, I was able to search by LCE.

PR #568, opened 2024-05-10
Fixes #27259 - Fix hostgroup-asset association

adamruzicka:
Because of the wrong cardinality, only first asset got removed on hostgroup destroy, leading to orphaned assets being left around.


ofedoren (approved):
Thanks, @adamruzicka and @pondrejk !

PR #570, opened 2024-05-15
Fixes #37464 - Support Zeitwerk loader

ofedoren:
~WIP~

Requires:
 - https://github.com/theforeman/foreman-tasks/pull/752

### DIFF
--- a/lib/foreman_openscap/engine.rb
+++ b/lib/foreman_openscap/engine.rb
@@ -5,15 +5,6 @@ module ForemanOpenscap
 
   class Engine < ::Rails::Engine
     engine_name 'foreman_openscap'
-    config.autoload_paths += Dir["#{config.root}/app/controllers/concerns"]
-    config.autoload_paths += Dir["#{config.root}/app/helpers/concerns"]
-    config.autoload_paths += Dir["#{config.root}/app/models/concerns"]
-    config.autoload_paths += Dir["#{config.root}/app/models"]
-    config.autoload_paths += Dir["#{config.root}/app/graphql"]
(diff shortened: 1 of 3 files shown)


lib/foreman_openscap/engine.rb
```diff
@@ -7,13 +7,6 @@ class Engine < ::Rails::Engine
     engine_name 'foreman_openscap'
     config.autoload_paths += Dir["#{config.root}/app/controllers/concerns"]
     config.autoload_paths += Dir["#{config.root}/app/helpers/concerns"]
```
ekohl, line 9: Why do these 2 remain, but you remove the others?

ofedoren: I think there is no reason to leave them, removed.

ofedoren:
Hmm, not sure why the tests from core are failing here... I had the same errors for plain Foreman, but they got fixed.

ekohl:
I think this may be a relevant failure, coming from STI with interfaces.

ofedoren:
@ekohl thanks, it was indeed. Moreover, thanks to that fail we prevented a hidden issue in REX caused by the same issue.

I'm afraid there might be still places we didn't catch by the test runs :/

ofedoren:
Ready to merge after https://github.com/theforeman/foreman/pull/10131 is merged, dropped the commit for testing with Foreman PR, the latest run was green:
Screenshot_20240903_143351

ofedoren:
tasks and REX are done, ansible is pending

ekohl (approved):
Note to self: foreman_ansible is a soft dependency so there's no version check that needs to be updated.

PR #572, opened 2024-05-20
Fixes #37487 - Refactor proxy selection

adamruzicka:


adamruzicka:
Alternatively, we could drop the version check completely, considering smart_proxy_openscap-0.5.0 was release in 2015 and therefore all currently supported versions have to have the required functionality.

ofedoren (approved):
Thanks, @adamruzicka, fixes the issue.

> Alternatively, we could drop the version check completely

I think if we remove the check, I'd say, we'd better make some (rather small) refactoring as well. It seems unnecessary for the fix itself though. What about fixing the issue this way, but remove the check with refactoring as the next step? On the other hand, if you don't want to create another ticket, which will be probably forgotten, we could do it here :)

ofedoren (approved):
Thanks, @adamruzicka, this should make things even cleaner. Also, still fixes the issue.

PR #574, opened 2024-07-02
Fixes #37619 - Add ouia-ids for remediation wizard

ofedoren:
@MariaAga, could you please take a look?

### DIFF
--- a/.github/workflows/js_tests.yml
+++ b/.github/workflows/js_tests.yml
@@ -12,6 +12,27 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  lint_custom:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        node-version: [14]
+    steps:
+      - uses: actions/checkout@v2
+      - name: Setup Node
+        uses: actions/setup-node@v1
+        with:
+          node-version:  ${{ matrix.node-version }}
+      - name: Npm install
+        run: |
+          npm install
+      - name: Run plugin linter
+        run: |
+          npm run lint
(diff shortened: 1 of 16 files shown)


webpack/components/ConfirmModal.js
```diff
@@ -11,6 +11,7 @@ const ConfirmModal = props => {
 
   const actions = [
     <Button
+      ouiaId={`oscap-conf-modal-${props.record.id}-confirm`}
```
MariaAga, line 14: `props.record` default is null. so this should be  `props.record?.id` or something else

.github/workflows/js_tests.yml
```diff
@@ -12,6 +12,27 @@ concurrency:
...
+      - name: Setup Node
+        uses: actions/setup-node@v1
+        with:
+          node-version:  ${{ matrix.node-version }}
+      - name: Npm install
+        run: |
+          npm install
+      - name: Run plugin linter
+        run: |
+          npm run lint
```
MariaAga, line 32: This already runs as part of `foreman_plugin_js`

MariaAga: maybe we should add it there, but thats a different pr :upside_down_face:

webpack/components/ConfirmModal.js
ofedoren: Fixed. Although, anything that is fixed here and is not related to the wizard should be removed, since those are leftovers from OVAL additions.

ofedoren:
Thanks, @MariaAga, updated.

MariaAga (approved):
lgtm!

PR #575, opened 2024-07-09
Fixes #37642 - Drop compatibility with smart_proxy_openscap < 0.6.1

ekohl:
If this was still needed, we could use capabilities and easily check this statically but version 0.6.1 was released in 2017 so by now it's safe to assume it's new enough.

Dropping the old code makes it more reliable and we have less code to maintain.

I must admit I've only grepped through the source and am not that familiar with the codebase, so it realistic that I missed something.


ekohl:
Yes, I missed something in the tests. Updated.

pondrejk (approved):

ekohl:
Updated the commit message to refer to a Redmine issue. No content change.

adamruzicka:
Thank you @ekohl & @pondrejk !

PR #577, opened 2024-07-23
Fixes #37677 - Extract RemWizard footer in a separate component

ofedoren:


adamruzicka:
Cross posting from another channel, this fixes the footer, but the close cross button in the header seems to still have autogenerated ouia id. Is that within our control?

ofedoren:
> Is that within our control?

Not quite. The wizard component we're using here has embedded and not changeable header subcomponent. We can hide stuff, but to add our, we'd need to write the wizard ourselves, I'm afraid.

Let's just ignore that since close cross button has the same function as cancel button.

MariaAga (approved):
LGTM

PR #578, opened 2024-07-26
Use documentation URL helper from core

ekohl:
Foreman 3.9.1+ has built in support to link to docs. This simplifies the code in this module.

There is some downstream branding consideration. Either https://github.com/RedHatSatellite/foreman_theme_satellite/blob/bf42fb119af723df5080e9b5d0f73fd8c483e017/lib/foreman_theme_satellite/documentation.rb#L67-L72 needs to be updated or https://github.com/RedHatSatellite/foreman_theme_satellite/pull/58 needs to be merged.


ofedoren (approved):
Thanks, @ekohl, LGTM, seems to work just fine

PR #579, opened 2024-08-20
Fixes #37755 - drop oval code leftovers in React

MariaAga:
This code is not used anywhere in the repo anymore


ofedoren:
Thanks, @MariaAga, I'm afraid there are more things to drop. Should we drop everything that's not used or we're dropping only "broken/bothering" parts?

MariaAga:
I found this while updating to pf5, and these were an effort to update, so deleting them looks easier.
I didn't see anything else thats unused, but I only checked a few things

adamruzicka (approved):
Some dropped parts are still better than no dropped parts, let's get this in

adamruzicka:
Thank you @MariaAga !

PR #583, opened 2024-10-31
Fixes #37961 - Mark ansible as job template requirement

adamruzicka:
This will prevent the job template from being seeded if ansible plugin is not available.

Soft requires https://github.com/theforeman/foreman_remote_execution/pull/923 . Without it, this won't do anything, but also won't break anything.

### DIFF
--- a/app/views/job_templates/run_openscap_remediation_-_ansible_default.erb
+++ b/app/views/job_templates/run_openscap_remediation_-_ansible_default.erb
@@ -16,6 +16,8 @@ template_inputs:
   description: Indicate wether the host should be rebooted after all the remediation
   input_type: user
   required: false
+require:
+- plugin: foreman_ansible
 %>
 ---
 - hosts: all


ofedoren (approved):
Thanks, @adamruzicka, this makes sense even without the soft dep.

Additional thanks for fixing my overlook :)

sbernhard:
Isn't it necessary to have something like https://github.com/Katello/katello/blob/master/db/seeds.d/75-job_templates.rb#L6 to make sure, that foreman_openscap can be used without installing foreman_ansible?

adamruzicka:
No, not anymore. After https://github.com/theforeman/foreman_remote_execution/pull/923 (and its dependency in forman) was merged, job template import should honor the dependencies declared in the templates themselves and skip those which don't have their requirements met

sbernhard:
ah, this is very nice. Would be cool to polish katello and other plugins to use this new approach.

PR #585, opened 2024-11-12
Fixes #38006 - Allow extension of list of packages

adamruzicka:
with scap contents


lib/foreman_openscap/bulk_upload.rb
```diff
@@ -7,21 +7,26 @@ def initialize
...
+    def package_installed?(package)
+      `rpm -qa | grep #{package}`.present?
     end
 
     def upload_from_scap_guide
-      unless scap_guide_installed?
-        @result.errors.push(_("Can't find scap-security-guide RPM, are you sure it is installed on your server?"))
+      package = security_guide_packages.find { |p| package_installed? p }
+      unless package
+        @result.errors.push(_("Can't find #{security_guide_packages.join(' or ')} RPM(s), are you sure it is installed on your server?"))
```
evgeni, line 25: The join should happen outside the translated string

ofedoren (commented):
Thanks, @adamruzicka, LGTM.

I've got only two comments:
 - Just a note: it kinda seems that there more changes than needed if we only care if any of these packages are installed. If we expect this list to be updated from even other plugins and the name of the package can differ from `scap-security-guide`, then it's fine. Or, if `theme_satellite` must enforce that `scap-sec-g-satellite` must have preference over the default one.
 - Requirement? :) I've tested with the counterpart, but it failed to load. It seems `BulkUpload` constant is not available during `foreman_theme_satellite` registration in `engine.rb`. That is probably I test on Rails 7 though. Long story short, could we move `bulk_upload.rb` into either `app/lib/foreman_openscap` or `app/services/foreman_openscap`? And just for complicity, remove https://github.com/theforeman/foreman_openscap/blob/master/app/controllers/api/v2/compliance/scap_contents_controller.rb#L1 and https://github.com/theforeman/foreman_openscap/blob/master/test/lib/foreman_openscap/bulk_upload_test.rb#L5 and https://github.com/theforeman/foreman_openscap/blob/master/lib/tasks/foreman_openscap_tasks.rake#L3

adamruzicka:
> If we expect this list to be updated from even other plugins

I don't really see that happening.

> Or, if theme_satellite must enforce that scap-sec-g-satellite must have preference over the default one.

Yeah, it must :/

> Requirement?

I'll take a look

ofedoren (changes requested):
Thanks, @adamruzicka, and please don't hate me, just a small nitpick. Otherwise, LGTM and seems to be working.

ofedoren (approved):
Thanks, @adamruzicka !

PR #590, opened 2025-01-28
Fixes #38164 - Update on bulk-upload from scap security guide

ofedoren:


app/services/foreman_openscap/bulk_upload.rb
```diff
@@ -48,12 +48,16 @@ def upload_from_files(files_array, from_scap_guide = false)
           next
         end
 
-        file = File.open(datastream, 'rb').read
+        file = File.read(datastream, mode: 'rb')
         digest = Digest::SHA2.hexdigest(datastream)
```
adamruzicka, line 52: Should we still keep calculating the digest from the filename?

ofedoren: I've left it just so we don't change behavior for non-standard content files.

For updates, it will be regenerated https://github.com/theforeman/foreman_openscap/blob/master/app/models/concerns/foreman_openscap/scap_file_content.rb#L11 based of file contents. I think we could change it to be generated based on contents right away on the first upload.

adamruzicka: > I think we could change it to be generated based on contents right away on the first upload.

That would make sense.

ofedoren: Now I'm thinking that it will change behavior for both cases (upload from custom files and scap guide): currently, on first upload it will generate digest from fullpath, stores it and then searches for existing db entry based on title and digest and if an entry is found it does nothing. By changing digest generation we will throw errors like "title already exists" on re-uploads.

Should we then just make it simple for all the cases: generate digest on upload in `before_validation` if it doesn't already exist (so we don't generate it twice on re-upload) and make already existing content updatable (as we want with scap guide) if titles match?

adamruzicka: > Should we then just make it simple for all the cases: generate digest on upload in before_validation if it doesn't already exist (so we don't generate it twice on re-upload) and make already existing content updatable (as we want with scap guide) if titles match?

That sounds like the best approach to me

ofedoren: Thanks, I've added these changes as a separate commit.

adamruzicka (approved):

adamruzicka:
Thank you @ofedoren !

PR #591, opened 2025-02-10
V9.0.5

Odilhao:


adamruzicka:
I have a hunch those requires might still be needed on pre-zeitwerk versions of Foreman. I'll have to double check that

adamruzicka (approved):

PR #592, opened 2025-02-11
Refs #37673 - change load to on("load"

MariaAga:
> Breaking change: .load(), .unload(), and .error() removed
These methods are shortcuts for event operations, but had several API limitations. The event .load() method conflicted with the ajax .load() method. The .error() method could not be used with window.onerror because of the way the DOM method is defined. If you need to attach events by these names, use the .on() method, e.g. change $("img").load(fn) to $("img").on("load", fn).


app/assets/javascripts/foreman_openscap/arf_reports.js
```diff
@@ -21,7 +21,7 @@ function buildArfModal(element, url) {
   var url = url + "?" + $.param({arf_report_ids: readFromCookie()});
   var title = $(element).attr('data-dialog-title');
   $('#confirmation-modal .modal-header h4').text(title);
-  $("#confirmation-modal .modal-body").load(url + " #content",
+  $("#confirmation-modal .modal-body"). on("load", (url + " #content",
```
MariaAga, line 24: probably needs a different format

adamlazik1 (approved):
LGTM and works.

app/assets/javascripts/foreman_openscap/arf_reports.js
ofedoren: These changes are gone from the latest commit. Even though, the new changes seem to fix this issue, shouldn't we still update these places as well?

MariaAga: no, I meant that the load in this file should not be on("load" , since it is calling a url and then doing a function, compared to the other one which wait to load  and then does a function

PR #593, opened 2025-02-20
i18n - pulling from tx

ekohl:


app/assets/javascripts/foreman_openscap/locale/ko/foreman_openscap.js
```diff
@@ -3,10 +3,10 @@
   "locale_data": {
     "foreman_openscap": {
       "": {
-        "Project-Id-Version": "foreman_openscap 10.1.0",
+        "Project-Id-Version": "foreman_openscap 10.2.0",
         "Report-Msgid-Bugs-To": "",
         "PO-Revision-Date": "2016-04-14 07:48+0000",
-        "Last-Translator": "FIRST AUTHOR <EMAIL@ADDRESS>, 2016",
+        "Last-Translator": "Ewoud Kohl van Wijngaarden <ewoud+transifex@kohlvanwijngaarden.nl>, 2025",
```
ofedoren, line 9: Didn't know you speak Korean :)

Just to ask: is there a reason we see spaces after some translated words like `Cancel`?

Looking at the diff, I see a lot of weird spaces and it's not only for Korean.

Not blocking, just wonder...

ekohl: > Didn't know you speak Korean :)

Neither did I :)

> Just to ask: is there a reason we see spaces after some translated words like `Cancel`?

I noticed those too. I didn't chase it too much now, but I wouldn't mind if someone took the effort to clean those up.

I'll leave it up to you if you want them fixed up before merging.

ofedoren: Just for the sake of easier synchronization, I'd leave that. It should be rather fixed in the "original" source.

ekohl: Oh yes, you're right about that.

ofedoren (approved):
Thanks, @ekohl !

PR #595, opened 2025-04-16
Fixes #38380 - remediations enable usr-overlay for bootc hosts

ianballou:
Adds a step to enable the usr-overlay for image mode hosts. Only works if the Katello plugin is installed.

Note: while we can enable the usr-overlay, for now, the package module will try using the atomic package manager, which will fail. Trying to edit playbooks sounds very risky, so we'll need to rely on the Ansible package module being fixed in the future (or the OpenSCAP playbooks).

Required: https://github.com/Katello/katello/pull/11374/

### DIFF
--- a/app/views/job_templates/run_openscap_remediation_-_ansible_default.erb
+++ b/app/views/job_templates/run_openscap_remediation_-_ansible_default.erb
@@ -22,22 +22,16 @@ require:
 ---
 - hosts: all
   tasks:
-    - name: Collect bootc status
-      shell:
-        cmd: 'bootc status --json'
-      register: bootc_status
-      ignore_errors: true
-    - name: Parse bootc status json
-      set_fact:
-        bootc_status_json: "{{ bootc_status.stdout | from_json }}"
-      when: bootc_status.rc == 0
+<%= indent(4) { snippet('check_bootc_status') } %>
     - name: Enable bootc overlay
       shell:
         cmd: 'bootc usr-overlay'
       register: out
       ignore_errors: true
-      when: bootc_status_json is defined and bootc_status_json['status']['booted'] != 'null'
+      when: is_bootc_host
     - debug: var=out
+      when: is_bootc_host
+<%= indent(4) { input('tasks') } -%>
 <% if truthy?(input('reboot')) %>
     - name: Reboot the machine
       reboot:
--- a/app/views/job_templates/run_openscap_remediation_-_ansible_default.erb
+++ b/app/views/job_templates/run_openscap_remediation_-_ansible_default.erb
@@ -22,7 +22,22 @@ require:
 ---
 - hosts: all
   tasks:
-<%= indent(4) { input('tasks') } -%>
+    - name: Collect bootc status
+      shell:
+        cmd: 'bootc status --json'
+      register: bootc_status
+      ignore_errors: true
+    - name: Parse bootc status json
+      set_fact:
+        bootc_status_json: "{{ bootc_status.stdout | from_json }}"
+      when: bootc_status.rc == 0
+    - name: Enable bootc overlay
+      shell:
+        cmd: 'bootc usr-overlay'
+      register: out
+      ignore_errors: true
+      when: bootc_status_json is defined and bootc_status_json['status']['booted'] != 'null'
+    - debug: var=out
 <% if truthy?(input('reboot')) %>
     - name: Reboot the machine
       reboot:
--- a/app/views/job_templates/run_openscap_remediation_-_script_default.erb
+++ b/app/views/job_templates/run_openscap_remediation_-_script_default.erb
@@ -17,6 +17,9 @@ template_inputs:
   input_type: user
   required: false
 %>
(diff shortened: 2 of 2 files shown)


ofedoren (commented):
I guess we need to adjust https://github.com/theforeman/foreman_openscap/blob/master/app/views/job_templates/run_openscap_remediation_-_script_default.erb as well?

ianballou:
This comment https://github.com/Katello/katello/pull/11374/files#r2048670891 also applies to this PR.

app/views/job_templates/run_openscap_remediation_-_script_default.erb
```diff
@@ -17,6 +17,9 @@ template_inputs:
   input_type: user
   required: false
 %>
+<% if @host.respond_to?(:image_mode_host?) && @host.image_mode_host? -%>
+bootc usr-overlay
```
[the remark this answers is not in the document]
ianballou: @ofedoren I think we might need to make a bootc host RFE with the project that holds the Ansible remediation playbooks. The request would be to detect if the host has a valid bootc status and use the dnf package manager if so.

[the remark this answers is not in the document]
ianballou: The alternative is to either edit the remediation playbooks (which sounds unsafe) or wait and see what happens with https://github.com/ansible/ansible/issues/84803

app/views/job_templates/run_openscap_remediation_-_ansible_default.erb
```diff
@@ -22,7 +22,14 @@ require:
 ---
 - hosts: all
   tasks:
-<%= indent(4) { input('tasks') } -%>
```
adamruzicka, line 25: This should be placed after the bootc stuff, right?

app/views/job_templates/run_openscap_remediation_-_ansible_default.erb
```diff
@@ -22,7 +22,14 @@ require:
...
+<%= snippet('check_bootc_status') %>
+    - name: Enable bootc overlay
+      shell:
+        cmd: 'bootc usr-overlay'
+      register: out
+      ignore_errors: true
+      when: is_bootc_host
+    - debug: var=out
```
adamruzicka, line 32: Would it make sense to print this only conditionally?

app/views/job_templates/run_openscap_remediation_-_script_default.erb
```diff
@@ -17,6 +17,9 @@ template_inputs:
   input_type: user
   required: false
 %>
+<% if @host.respond_to?(:image_mode_host?) && @host.image_mode_host? -%>
```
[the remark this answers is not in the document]
adamruzicka: > but I can see people working with image mode machines on vanilla Foreman.

In theory they could, but without katello we don't have any means of distinguishing image mode and package mode hosts, right?

adamruzicka (changes requested):

app/views/job_templates/run_openscap_remediation_-_script_default.erb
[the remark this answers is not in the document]
ianballou: Exactly, we'd need to use bootc status otherwise like in Ansible (which wasn't exactly done for vanilla Foreman compatibility reasons)

app/views/job_templates/run_openscap_remediation_-_ansible_default.erb
ianballou: Perhaps, I included it because the Foreman Ansible shell command template includes the same debug output.
What would it look like as a conditional?

app/views/job_templates/run_openscap_remediation_-_ansible_default.erb
ianballou: All set now.

app/views/job_templates/run_openscap_remediation_-_ansible_default.erb
```diff
@@ -22,6 +22,14 @@ require:
...
 - hosts: all
   tasks:
+<%= indent(4) { snippet('check_bootc_status') } %>
+    - name: Enable bootc overlay
+      shell:
+        cmd: 'bootc usr-overlay'
+      register: out
+      ignore_errors: true
+      when: is_bootc_host
+      - debug: var=out
```
adamruzicka, line 32: The indentation looks off
```suggestion
    - debug: var=out
```

adamruzicka, line 32: With indentation and conditional it would be
```suggestion
    - debug: var=out
      when: is_bootc_host
```

ianballou: Thanks, I haven't created a setup to actually test this myself

ianballou:
The indentation should be fixed now.

adamruzicka (approved):
LGTM

ofedoren (approved):
Thanks, @ianballou and @adamruzicka !

PR #596, opened 2025-06-16
use only @theforeman/actions for JS tests

MariaAga:
lint_custom runs as part of the generic test action

### DIFF
--- a/.github/workflows/js_tests.yml
+++ b/.github/workflows/js_tests.yml
@@ -12,24 +12,6 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  lint_custom:
-    runs-on: ubuntu-latest
-    strategy:
-      fail-fast: false
-      matrix:
-        node-version: [14]
-    steps:
-      - uses: actions/checkout@v2
-      - name: Setup Node
-        uses: actions/setup-node@v1
-        with:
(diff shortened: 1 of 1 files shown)


ofedoren:
Could you add a commit, which uses https://github.com/theforeman/foreman/pull/10572, to confirm it fails here?

MariaAga:
Not sure why it didnt work ):

.github/workflows/js_tests.yml
```diff
@@ -12,26 +12,10 @@ concurrency:
...
-      - name: Run custom plugin linter
-        run: |
-          npm run lint:custom
   test:
     name: JavaScript
     uses: theforeman/actions/.github/workflows/foreman_plugin_js.yml@v0
     with:
       plugin: foreman_openscap
+      foreman_version: gh-actions
+      foreman_repository: MariaAga/foreman
```
ofedoren, line 21: Maybe this will work? I mean, there is almost no difference, I just always used this:
```suggestion
      foreman_version: refs/pull/10572/head
```

MariaAga: It does look more like what I saw in the past, but couldnt find it written anywhere 
Thanks

ofedoren (approved):
Thanks, @MariaAga !

Please don't forget to remove test commits before merge.

PR #600, opened 2025-10-08
i18n - pulling from tx

adamlazik1:


adamlazik1:
Yours looks better, although I don't know what I did differently to get botched output.

PR #607, opened 2026-03-13
Fixes #39148 - Take name from host directly

adamruzicka:
without taking a detour through asset when building a filename when client tries to download an arf report in html or csv formats.

The arf_report-asset association may not exist if the host gets its policy from a hostgroup.

Fixes 9b4995a


adamlazik1 (commented):
Looks good, one technical question

adamlazik1 (approved):

adamlazik1:
Thanks @adamruzicka !

PR #608, opened 2026-03-16
Fixes #39148 - Take name from host directly

adamruzicka:
without taking a detour through asset when building a filename when
client tries to download an arf report in html or csv formats.

The arf_report-asset association may not exist if the host gets its
policy from a hostgroup.

Fixes 9b4995a

(cherry picked from commit 1545c0a264d17e12a81afb877810e610569fde5e) (https://github.com/theforeman/foreman_openscap/pull/607)


adamlazik1 (approved):
LGTM

adamlazik1:
Thanks @adamruzicka !

PR #609, opened 2026-04-27
Fixes #39265 - Fix dependency issues for foreman_openscap

ofedoren:


Gemfile
```diff
@@ -1,5 +1,7 @@
 source "http://rubygems.org"
 
-gem 'theforeman-rubocop', '~> 0.1.1', groups: %i[development rubocop]
-
 gemspec
+
+gem 'rake'
```
adamruzicka, line 5: Why do we even directly depend on rake?

Gemfile
```diff
@@ -1,5 +1,7 @@
...
+gem 'rdoc'
```
adamruzicka: I'd lean towards dropping it altogether

Gemfile
ofedoren: You won't like it: without this any `bundle exec rake` will fail with `can't find executable rake for gem rake. rake is not currently included in the bundle, perhaps you meant to add it to your Gemfile?`

UPD: In other words, without explicit dependency on `rake`, `Rakefile` is useless for plugin-only context, e.g. `cd foreman_openscap ; bundle install ; bundle exec rake -T` fails.

adamruzicka: Well, yes, but are there any rake tasks that we should still be using directly from the plugin repo?

ofedoren: ~AFAIU most of the tasks we run either in CI or within Foreman context. I personally didn't run anything from this Rakefile in a... never?~ I'm a big liar: https://github.com/theforeman/foreman_openscap/commit/afb50064f6050b99f0632f31847619cbe4b43253. ~But I can't remember why I've added that rake task though...~ That rake task just because if there's a Rakefile then the CI will run it via `bundle exec rake` instead of `bundle exec rubocop`.

UPD: We can leave rake dependency just to leave Rakefile in case there will be a need in one, but looking at it currently, I don't think we much need that. WDYT, leaving or dropping?

adamruzicka: If it makes you sleep better, I don't really remember it either. Well, nevermind then, let's leave it in for the time being

ofedoren: Sorry for spam-edit comments, I never expect someone to react immediately :D

Anyway, leaving Rakefile for time being.

Gemfile
ofedoren: Dropped in a separate commit

adamruzicka (approved):

adamruzicka:
Thank you @ofedoren !

PR #612, opened 2026-05-29
Fixes #39380 - Update line chart to PatternFly5

tlabaj:
Fixes: #39380

This PR updates the Line Chart to use PF5 components.

Before:
linechart_before

After:
Screenshot 2026-06-03 at 11 24 23 AM

### DIFF
--- a/app/assets/javascripts/foreman_openscap/scap_hosts_show.js
+++ b/app/assets/javascripts/foreman_openscap/scap_hosts_show.js
@@ -1,4 +1,0 @@
-$(function() {
-  // Not sure about this ugly hack.
-  $('.col-md-4 .stats-well').height($('.col-md-8 .stats-well').height());
-});
--- a/webpack/components/LineChart/LineChartHelpers.js
+++ b/webpack/components/LineChart/LineChartHelpers.js
@@ -1,0 +1,180 @@
+/** Backend sends [label, values, color]; legacy Foreman charts used spread columns. */
+import { chart_color_black_500 as chartColorBlack500 } from '@patternfly/react-tokens';
+
+const getColumnValues = col => {
+  if (Array.isArray(col[1])) return col[1];
+
+  const values = col.slice(1);
+  const last = values[values.length - 1];
+
+  if (
+    values.length > 1 &&
+    (last == null || (typeof last === 'string' && last.startsWith('#')))
+  ) {
+    return values.slice(0, -1);
+  }
+
+  return values;
+};
+
+const getSeriesColor = col => {
+  if (typeof col[2] === 'string' && col[2].startsWith('#')) {
+    return col[2];
+  }
+
+  const last = col[col.length - 1];
+  if (typeof last === 'string' && last.startsWith('#')) {
+    return last;
+  }
+
+  return undefined;
+};
+
+const toMs = val => {
+  const n = Number(val);
+  return n >= 1e12 ? n : n * 1000;
+};
+
+/** Process raw backend data into chart series for PatternFly line charts. */
+export const processChartData = (data, xAxisDataLabel, config) => {
+  if (!data || data.length === 0) return null;
+
+  if (config === 'timeseries') {
+    const timeColumn = data.find(col => col[0] === xAxisDataLabel);
+    if (!timeColumn) return null;
+
+    const xValues = getColumnValues(timeColumn).map(t => toMs(t));
+
+    const series = data
+      .filter(col => col[0] !== xAxisDataLabel)
+      .map(col => {
+        const values = getColumnValues(col);
+        return {
+          name: col[0],
+          color: getSeriesColor(col),
+          data: values.map((value, index) => ({
+            x: new Date(xValues[index]),
+            y: value ?? 0,
+            name: col[0],
+          })),
+        };
+      });
+
+    return series.length > 0 ? series : null;
+  }
+
+  const series = data.map(col => {
+    const values = getColumnValues(col);
+    return {
+      name: col[0],
+      color: getSeriesColor(col),
+      data: values.map((value, index) => ({
+        x: index + 1,
+        y: value ?? 0,
+        name: col[0],
+      })),
+    };
+  });
+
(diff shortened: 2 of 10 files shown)


webpack/lineChartRegistry.js
```diff
@@ -0,0 +1,27 @@
...
+export const registerOpenscapLineChart = () => {
+  componentRegistry.registry[COMPONENT_NAME] = {
+    type: LineChart,
+    store: true,
+    data: true,
+  };
+
+  remountComponents();
+};
+
```
MariaAga: Why is it being registered like this?

app/assets/javascripts/foreman_openscap/scap_hosts_show.js
Lukshio, line 1: We are trying to get rid of JQuery custom hooks, this logic should be replaced by CSS.
https://github.com/theforeman/foreman/blob/develop/developer_docs/legacy-js.asciidoc

webpack/components/LineChart/LineChartHelpers.js
Lukshio, line 1: This helper function are already in foreman core in `AreaChartLegend` so we should reuse that instead of duplicating it. We can open new PR in core that will move the functions to some general chart helpers.

```
  getXAxisTickValues,
  formatAxisTick,
  formatYAxisTick,
  formatTooltipTitle,
  XAxisTickLabel,
```

webpack/components/LineChart/index.js
```diff
@@ -0,0 +1,237 @@
...
+                          }
+                          return null;
+                        },
+                      },
+                    ],
+                  },
+                },
+              ]
+            : undefined
+        }
```
Lukshio: This can be extracted to some const

webpack/components/LineChart/index.js
```diff
@@ -0,0 +1,237 @@
...
+  ChartAxis,
+  ChartGroup,
+  ChartLine,
+  ChartLegendTooltip,
+  ChartThemeColor,
+  createContainer,
+} from '@patternfly/react-charts';
+
+import { translate as __ } from 'foremanReact/common/I18n';
+import MessageBox from 'foremanReact/components/common/MessageBox';
```
Lukshio: Message box will be deprecated, use `EmptyState` component instead

Lukshio (changes requested):
Hi, I know it's WIP, but just first round of review.

Lukshio:
Hi, please commit new changes in new commits.

tlabaj:
> Hi, please commit new changes in new commits.

Sure, I will do that going forward.

webpack/components/LineChart/index.js
```diff
@@ -0,0 +1,317 @@
...
+  formatAxisTick,
+  formatTooltipTitle,
+  formatYAxisTick,
+  getLegendEvents,
+  getSeriesOpacity,
+  getXAxisTickValues,
+  InteractiveLegendLabel,
+  InteractiveLegendSymbol,
+  XAxisTickLabel,
+} from 'foremanReact/components/common/charts/AreaChart/AreaChartLegend';
```
Lukshio: Can you please open new PR in foreman core and move all imports to new `/chars/helpers/LegendHelpers.js` so it can be easily reused later? So it won't be matched to specific chart. The `AreaChartLegend` will then use new helper.

Path update will be needed after the change.

Lukshio (commented):
Code changes looks good

webpack/components/LineChart/index.js
tlabaj: @Lukshio Here is the Foreman PR https://github.com/theforeman/foreman/pull/11010

Lukshio:
@tlabaj ForemanCore PR is merged, you can edit import paths

ofedoren:
> @tlabaj ForemanCore PR is merged, you can edit import paths

Friendly reminder: If a dependency in Foreman is required and was merged, please update https://github.com/theforeman/foreman_openscap/blob/master/lib/foreman_openscap/engine.rb#L43 accordingly

Lukshio (commented):
Code changes looks good, please update the `engine.rb` to new foreman version, and fix CI failures

tlabaj:
> Code changes looks good, please update the `engine.rb` to new foreman version, and fix CI failures

I updated the `engine.rb` file.  Is it Ok to squash commits now to fix CI?

Lukshio:
> I updated the `engine.rb` file. Is it Ok to squash commits now to fix CI?

@tlabaj Redmine is OK, but linter erros has be fixed.

ofedoren:
If it's going to be merged after linter fix, please squash before or on merge.

ofedoren:
Not a blocker, but a question: I've seen Foreman core already has a LineChart component (https://github.com/theforeman/foreman/blob/develop/webpack/assets/javascripts/react_app/components/common/charts/LineChart/index.js#L19), but this PR adds another one specifically for openscap. Why's that? Wouldn't it be better to migrate the core's one to PF5 and then re-use it here?..

MariaAga:
@ofedoren We prefer using PF directly and through wrappers, and matching each component use to the format of data it receives. From what I saw, trying to create wrappers that will match different types of needs and data, resulted in complicated code to re-use, read and update. (disclaimer: I did a quick look at the code only).
And we will be deprecating the foreman core line chart

pnovotny:
QA-wise LGTM. 
PR has been verified on top of Satellite stream snap. 194.

_Results_
Before:
compliance-hosts-before

After:
compliance-hosts-after

One small nitpick (non-blocking): while using float for the Y-axis is fine-ish (WRT scaling and such), using it in the tooltip with Passed/Failed/Othered rules looks a bit strange, as a rule is non-divisible.

Lukshio:
@tlabaj could you please change the float numbers?

pnovotny:
Re-verified with the latest PackIt build. The float numbers are fixed. Everything else looks also fine :+1: 

line-chart-float-fixed

Lukshio (approved):
dev ack

PR #613, opened 2026-06-23
Manage compliance status if policy only assigned to hostgroup

sbernhard:
If a host doesn't have policies assigned directly, but only via hostgroup, it still should re-calculate the compliance status.


app/models/foreman_openscap/compliance_status.rb
```diff
@@ -43,11 +43,15 @@ def to_global(options = {})
...
-      host.policies.present?
+      host&.policies.present? || host&.hostgroup&.policies.present?
     end
 
     def to_status(options = {})
-      latest_reports = host.policies.map { |p| host.last_report_for_policy p }.flatten
+      policies = (
+        host.policies.to_a +
+        host.hostgroup&.policies.to_a
+      ).uniq
```
adamruzicka: Wouldn't this be a better fit?
```suggestion
      policies = host.combined_policies
```

sbernhard: yes, thanks.

app/models/foreman_openscap/compliance_status.rb
```diff
@@ -43,11 +43,11 @@ def to_global(options = {})
     def relevant?(options = {})
       # May fail host status during migration
       return false unless ForemanOpenscap::Asset.table_exists?
-      host.policies.present?
+      host&.policies.present? || host&.hostgroup&.policies.present?
```
adamruzicka: We may need to also use combined policies here, iirc `host&.hostgroup&.policies` doesn't take into account policies that the hostgroup might be inheriting from its parent(s)

adamruzicka (commented):
Could we also get a redmine?

app/models/foreman_openscap/compliance_status.rb
sbernhard: tested with combined_policies. did get a true/false, too

app/models/foreman_openscap/compliance_status.rb
```diff
@@ -43,11 +43,11 @@ def to_global(options = {})
     def relevant?(options = {})
       # May fail host status during migration
       return false unless ForemanOpenscap::Asset.table_exists?
-      host.policies.present?
+      host.combined_policies.present?
```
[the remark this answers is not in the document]
dosas: I'm with the AI on this one ;)

sbernhard:
@adamruzicka good to merge?

adamruzicka (approved):
lgtm

adamruzicka:
Thank you @sbernhard !

PR #614, opened 2026-06-30
Refs #39448 - improve performance on host index page

sbernhard:
During testing, we found out multiple issues so that the performance of the host index page is really bad.
Before this commit, using '/api/hosts?include_permissions=false&search=' took about 1min 30seconds. Simply returning "false" for "relevant?" decreased the access to 6seconds. With this commit, we are back at 9seconds - which is acceptable hopefully.

Replace call to ".to_status" with ".status" to use the already calculated value of the complianace report.

host_has_hostgroup_policies? is pretty fast because it answers a yes/no question with a cheap EXISTS query instead of loading and deduplicating all applicable policies in Ruby.

### DIFF
--- a/app/models/concerns/foreman_openscap/host_extensions.rb
+++ b/app/models/concerns/foreman_openscap/host_extensions.rb
@@ -129,12 +129,12 @@ module ForemanOpenscap
       report_scope
     end
 
-    def compliance_status(options = {})
-      @compliance_status ||= get_status(ForemanOpenscap::ComplianceStatus).to_status(options)
+    def compliance_status(_options = {})
+      @compliance_status ||= get_status(ForemanOpenscap::ComplianceStatus).status
     end
 
-    def compliance_status_label(options = {})
-      @compliance_status_label ||= get_status(ForemanOpenscap::ComplianceStatus).to_label(options)
+    def compliance_status_label(_options = {})
+      @compliance_status_label ||= get_status(ForemanOpenscap::ComplianceStatus).to_label
     end
 
     def openscap_proxy_in_taxonomy
--- a/app/models/foreman_openscap/asset.rb
+++ b/app/models/foreman_openscap/asset.rb
@@ -1,7 +1,9 @@
 module ForemanOpenscap
   class Asset < ApplicationRecord
     has_many :asset_policies, :dependent => :delete_all
-    has_many :policies, :through => :asset_policies
+    has_many :policies, :through => :asset_policies,
+                        :after_add => :reset_compliance_status,
+                        :after_remove => :reset_compliance_status
     belongs_to :assetable, :polymorphic => true
 
     scope :hosts, lambda { where(:assetable_type => 'Host::Base') }
@@ -19,6 +21,24 @@ module ForemanOpenscap
     end
 
     private
+
+    def reset_compliance_status(_policy)
+      ForemanOpenscap::ComplianceStatusResetter.to_inconclusive(host_ids_reached_by_asset)
+    end
+
+    def host_ids_reached_by_asset
+      case assetable_type
+      when 'Host::Base'
+        [assetable_id]
+      when 'Hostgroup'
+        hostgroup = Hostgroup.find_by(:id => assetable_id)
+        return [] unless hostgroup
+
+        Host.where(:hostgroup_id => hostgroup.subtree_ids).pluck(:id)
+      else
+        []
+      end
+    end
 
     def fetch_asset(type)
       assetable if assetable_type == type
--- a/app/models/foreman_openscap/compliance_status.rb
+++ b/app/models/foreman_openscap/compliance_status.rb
@@ -16,8 +16,8 @@ module ForemanOpenscap
       "#{ArfReport::BIT_NUM * ArfReport::METRIC.index(status)} & #{ArfReport::MAX}"
     end
 
-    def to_label(options = {})
-      case to_status
+    def to_label(_options = {})
+      case status
       when COMPLIANT
         N_('Compliant')
       when INCONCLUSIVE
@@ -29,8 +29,8 @@ module ForemanOpenscap
       end
     end
 
-    def to_global(options = {})
-      case to_status
+    def to_global(_options = {})
+      case status
       when COMPLIANT
         ::HostStatus::Global::OK
       when INCONCLUSIVE
@@ -40,17 +40,29 @@ module ForemanOpenscap
       end
     end
 
-    def relevant?(options = {})
+    def relevant?(_options = {})
       # May fail host status during migration
       return false unless ForemanOpenscap::Asset.table_exists?
-      host.combined_policies.present?
+      host.policies.any? || host_has_hostgroup_policies?
     end
 
-    def to_status(options = {})
+    def to_status(_options = {})
       latest_reports = host.combined_policies.flat_map { |p| host.last_report_for_policy p }
       return INCOMPLIANT if latest_reports.any?(&:failed?)
       return INCONCLUSIVE if latest_reports.any?(&:othered?)
       COMPLIANT
+    end
+
+    private
+
+    def host_has_hostgroup_policies?
+      return false if host.hostgroup_id.blank?
+
+      ids = [host.hostgroup_id] + host.hostgroup.ancestor_ids
+      ForemanOpenscap::Policy.joins(:assets)
+                             .where('foreman_openscap_assets.assetable_type' => 'Hostgroup',
+                                    'foreman_openscap_assets.assetable_id' => ids)
+                             .exists?
     end
   end
 end
--- a/app/models/foreman_openscap/policy.rb
+++ b/app/models/foreman_openscap/policy.rb
@@ -42,6 +42,7 @@ module ForemanOpenscap
     validate :valid_tailoring, :valid_tailoring_profile, :no_mixed_deployments
     validate :valid_cron_line, :valid_weekday, :valid_day_of_month, :if => Proc.new { |policy| policy.should_validate?('Schedule') }
     after_save :assign_policy_to_hostgroups
+    after_save :reset_changed_assignment_hosts_to_inconclusive
     # before_destroy - ensure that the policy has no hostgroups, or classes
 
     default_scope do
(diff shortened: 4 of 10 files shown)


app/models/concerns/foreman_openscap/host_extensions.rb
```diff
@@ -130,7 +130,7 @@ def reports_for_policy(policy, limit = nil)
     end
 
     def compliance_status(options = {})
-      @compliance_status ||= get_status(ForemanOpenscap::ComplianceStatus).to_status(options)
+      @compliance_status ||= get_status(ForemanOpenscap::ComplianceStatus).status
```
nadjaheitmann, line 133: I cannot say much about performance impact, but I see that you are not using the options parameter anymore. So either it should be removed from the parameter list or it should be used somehow.

adamruzicka:
I like the general idea, it has some other benefits over just being faster (like search finally working reliably). However this sort of deviates from the pattern we generally use for statuses and makes the whole thing much more prone to showing stale data. For example we don't refresh the status when a policy is assigned/unassigned to the host (or its hostgroup or its ancestors). So either we should keep building the values dynamically or make sure to refresh the sub-status whenever any change that could affect its value happens.

sbernhard:
> For example we don't refresh the status when a policy is assigned/unassigned to the host (or its hostgroup or its ancestors)

In katello, like https://github.com/Katello/katello/blob/master/app/models/katello/errata_status.rb#L25 it is done the same way. Actually, the compliance_status need to be updated if there is a new report. If another policy is set, its "nice" that the status is set ASAP to "not compliant" but it become really meanigful, if there is a report including this policy.

adamruzicka:
> but it become really meanigful, if there is a report including this policy.

I was thinking in slightly different direction - if a host was incompliant with its policy and that policy was replaced with some other, I'd expect the status to turn green immediately, even if (or maybe better "because") it has no reports with the new policy

jtruestedt:
So if you have a host and a report, which tells the host is compliant/incompliant and you change the assigned policy, you cannot know if it is still compliant/incompliant until a report is there. So from my point of view the status should be "unknown" (warning instead of ok/error) or even empty. 
Imagine you have a host which is incompliant, so you change the policy and it will turn to green and on the other side you make sure that no new report is sent - then the host will be compliant forever by changing the policy. Same counts somehow for having to switch to "incompliant" but this shows at least that some action is needed. 

Just to add the user-perspective without knowing what the code does.

app/models/concerns/foreman_openscap/host_extensions.rb
sbernhard: I had a look again at various plugins. These "#{status_class.humanized_name}_status_label" are used in different plugins and all have "options". I changed it to "_options"

sbernhard:
Another thought but not yet verified. The relevant? method checks for, if policies are set. Maybe we should instead check, if there is a report? maybe this is faster.

app/models/foreman_openscap/compliance_status.rb
```diff
@@ -40,17 +40,27 @@ def to_global(options = {})
...
       latest_reports = host.combined_policies.flat_map { |p| host.last_report_for_policy p }
       return INCOMPLIANT if latest_reports.any?(&:failed?)
       return INCONCLUSIVE if latest_reports.any?(&:othered?)
       COMPLIANT
     end
+
+    private
+
+    def host_has_hostgroup_policies?
+      ids = [host.hostgroup_id] + host.hostgroup.ancestor_ids
```
m-bucher, line 61: Would make the check in `relevant?()` shorter:
```
host.policies.any? || host_has_hostgroup_policies?
```

```suggestion
    def host_has_hostgroup_policies?
      return false if host.hostgroup_id.blank?
      
      ids = [host.hostgroup_id] + host.hostgroup.ancestor_ids
```

sbernhard: This is done by https://github.com/theforeman/foreman_openscap/pull/614/changes#diff-fe79f8a02bc1180697196bb0f30bf6b048a2237325a79c411f87f2de5191b73eR46 before?

sbernhard: Thanks. Took your solution.

ShimShtein:
From looking at the PR, it seems correct. The whole idea is not to recalculate the status on the fly (which can get really expensive).
I do miss in this PR the places when the status should actually be recalculated. I would expect to see a recalculation on status report. Also I would like to see recalculation on host properties change that are relevant for status calculations. For example in rh_cloud we recalculate the status when a certain action happens: https://github.com/theforeman/foreman_rh_cloud/blob/1faeed9fefb7533af906f8d01ac96c69d7a72d85/lib/inventory_sync/async/inventory_full_sync.rb#L46
Notice that if the action is not user-facing, we don't care about performance that much. I have seen the talks about policy change - this is a good example where we can fire off an async action that will invalidate statuses for all the relevant hosts.

sbernhard:
> I do miss in this PR the places when the status should actually be recalculated. I would expect to see a recalculation on status report. 

adjusting the compliance status is done by https://github.com/theforeman/foreman_openscap/blob/a276f0b764fb1142df8c9fb9925b023703cc6fd1/app/controllers/api/v2/compliance/arf_reports_controller.rb#L53 , isn't it?

I 'm currently again looking if it would be easily possible to set the status to inconclusive if policies are added/removed. 

Regarding backgroup job, without foreman-tasks this might be a challenge, too :)

sbernhard:
I have changed this to draft again. 
I have added https://github.com/theforeman/foreman_openscap/pull/614/commits/e31ed0c07bf4863c8099e331c98670a366bdd419 - but not yet tested. All of this is required, to make sure, that the compliance status for all hosts is set to 'INCONCLUSIVE' for which a policy was added / removed - no matter if directly to host, hostgroup or parent(s) host group. 
Is it worth? @ShimShtein / @adamruzicka  / @m-bucher

ShimShtein:
Did a very high-level pass on the inconclusive commit. Looks quite OK to me. We may be able to improve it a bit, but overall it's OK.
As to whether it's worth it - a good question. It does make our DB much more consistent.

sbernhard:
> Looks quite OK to me. We may be able to improve it a bit, but overall it's OK. 

can you let me know, what you want to improve?

>As to whether it's worth it - a good question. It does make our DB much more consistent.

I have tested it and it looks like this. From the performance perspecitve, its OK to reset the state.
What I have tested:
- using the old all-hosts page -> set compliance policy for a host -> state is set to INCONCLUSIVE. Run compliance report -> state is set accordingly
- using the old all-hosts page -> unset compliance policy for a host which has a hostgroup policy, too -> state is set to INCONCLUSIVE. Run compliance report -> state is afterward set accordingly
- using the old all-hosts page -> unset compliance policy for a host without hostgroup policy-> state is set to N/A because the host and its hostgroup does not have a policy (=> this is correct)

- Hosts>Compliance>Policies - set policy for a hostgroup -> state is set to INCONCLUSIVE for the hosts of this group and inherited groups. Run compliance report -> state is afterward set accordingly
- Hosts>Compliance>Policies - unset policy for a hostgroup -> state is set to INCONCLUSIVE  for all hosts of this group and inherited groups if they have direct host policies or other hostgroup policies. Run compliance report -> state is afterward set accordingly.
-  Hosts>Compliance>Policies - unset policy for a hostgroup -> state is set to N/A  for the hosts of this group and inherited groups if they don't have direct host policy or other host group policies.

=> this PR works!


Additionally, I want to notice again, that its currently not possible on the new all-hosts page to set policies to hosts - only host group because the bulk actions on the new all hosts page to set / unset compliance policy for a host don't exist:
https://community.theforeman.org/t/deprecating-and-removing-the-legacy-content-hosts-ui/42729/13

Actually, for a plugin like "openscap" it would be super good to have a host details card in which it is possible to a) see  which policies are set and b) set a host policy and c) see the compliance status and maybe link to last report - this wouldn't be that hard to implement....

app/models/foreman_openscap/policy.rb
```diff
@@ -251,12 +256,47 @@ def profile_for_scan
...
+      @compliance_status_reset_host_ids ||= []
+      @compliance_status_reset_host_ids.concat(host_ids)
+    end
+
+    def host_assignment_changed?(before_host_ids, after_host_ids)
+      before_host_ids.to_set != after_host_ids.to_set
+    end
+
+    def reset_changed_assignment_hosts_to_inconclusive
+      host_ids = Array(@compliance_status_reset_host_ids).compact.uniq
```
m-bucher, line 294: This could be a getter-method:

```suggestion
    def compliance_status_reset_host_ids
      Array(@compliance_status_reset_host_ids).compact.uniq
    end
    
    def reset_changed_assignment_hosts_to_inconclusive
      host_ids = compliance_status_reset_host_ids
```
But as long as we only use the value once, it does not really make a difference.

sbernhard: .. or it's another function call which isn't really necessary ;-)

sbernhard:
@adamruzicka  how to continue with this?

sbernhard:
How should we continue with this @adamruzicka ?

ShimShtein:
Made another round of review. Looks fine to me. Let's wait for Adam's review.

app/services/foreman_openscap/compliance_status_resetter.rb
```diff
@@ -0,0 +1,20 @@
...
+      rows = host_ids.map do |host_id|
+        {
+          :host_id => host_id,
+          :type => ComplianceStatus.to_s,
+          :status => ComplianceStatus::INCONCLUSIVE,
+          :reported_at => now
+        }
+      end
+
+      ComplianceStatus.upsert_all(rows, :unique_by => %i[type host_id], :update_only => %i[status reported_at])
```
adamruzicka, line 17: This updates the sub-status, but this change is not propagated upwards to the global status. Is that intentional?

app/models/foreman_openscap/asset.rb
```diff
@@ -1,7 +1,9 @@
 module ForemanOpenscap
   class Asset < ApplicationRecord
     has_many :asset_policies, :dependent => :delete_all
-    has_many :policies, :through => :asset_policies
+    has_many :policies, :through => :asset_policies,
+                        :after_add => :reset_compliance_status,
+                        :after_remove => :reset_compliance_status
```
adamruzicka, line 6: This (and policy L45) would still miss things like hostgroup reparenting and re-assigning a host to a different hostgroup, right?

PR #615, opened 2026-07-07
Add OpenSCAP compliance card to host details

sbernhard:
The compliance card has
- Show last state
- Show Rule result of last report
- Show date of last report and link to last report
- Show how many policies. If you click on the Policy count you see all used policies and the source (host, hostgroup). Additionally, its possible to set / unset directly to host.

Card:
image

Modal view to assign / unassign policies. It shows host and host group assignments and allows to assign / unassign direct host assignments:
image

PR #616, opened 2026-07-16
Fixes #39532 - All hosts - add action - bulk change proxy

Lukshio:
Users can assign an OpenSCAP Proxy to a host. This can be done individually for a host via the host edit form, but also in bulk via the legacy host overview UI (see screenshot). The new UI currently doesn't support this.

Option can be found on AllHosts page -> select hosts -> kebab menu -> Change association -> OpenSCAP Proxy


webpack/components/HostsIndex/BulkActions/changeOpenscapProxy/BulkChangeOpenscapProxyModal.js
```diff
@@ -0,0 +1,275 @@
...
+    );
+  };
+
+  const descriptionText = selectAllHostsMode ? (
+    <>
+      {__('Assign OpenSCAP capsule for ')}
+      <strong>{__('ALL selected hosts')}</strong>
+      <br />
+      {__(
+        '. This will change previous capsule assignments on the selected hosts.'
```
MariaAga: Image

the `.` should be before the `<br`

webpack/components/HostsIndex/BulkActions/changeOpenscapProxy/BulkChangeOpenscapProxyModal.js
MariaAga, line 1: Needs tests

webpack/components/HostsIndex/ChangeOpenscapProxyAction.js
MariaAga, line 1: Needs tests

webpack/components/HostsIndex/BulkActions/changeOpenscapProxy/BulkChangeOpenscapProxyModal.js
```diff
@@ -0,0 +1,275 @@
...
+    </Button>,
+  ];
+
+  return (
+    <Modal
+      isOpen={isOpen}
+      onClose={handleModalClose}
+      onEscapePress={handleModalClose}
+      title={__('Assign OpenSCAP Capsule')}
+      width={650}
```
MariaAga: should just be variant=medium

webpack/components/HostsIndex/BulkActions/changeOpenscapProxy/BulkChangeOpenscapProxyModal.js
```diff
@@ -0,0 +1,275 @@
...
+        handleError,
+        errorToast: error =>
+          error?.response?.data?.error?.message || __('Error'),
+        params: requestBody,
+      })
+    );
+  };
+
+  const descriptionText = selectAllHostsMode ? (
+    <>
```
MariaAga, line 106: `FormattedMessage` might be helpful here

webpack/components/HostsIndex/BulkActions/changeOpenscapProxy/BulkChangeOpenscapProxyModal.js
```diff
@@ -0,0 +1,275 @@
...
+                      onOpenChange={isSelectOpen =>
+                        setProxySelectOpen(isSelectOpen)
+                      }
+                      ouiaId="bulk-change-openscap-capsule-select"
+                      toggle={toggleRef => (
+                        <MenuToggle
+                          ref={toggleRef}
+                          onClick={handleToggleClick}
+                          isExpanded={proxySelectOpen}
+                          style={{ width: '100%' }}
```
MariaAga: `isFullWidth` is preferred

app/controllers/concerns/foreman_openscap/api/v2/hosts_bulk_actions_controller_extensions.rb
MariaAga, line 1: Success message doesn't include host count (inconsistent with core bulk actions)

webpack/components/HostsIndex/BulkActions/changeOpenscapProxy/BulkChangeOpenscapProxyModal.js
```diff
@@ -0,0 +1,275 @@
...
+    >
+      <Stack hasGutter>
+        <StackItem>{descriptionText}</StackItem>
+        {proxyStatus === STATUS.RESOLVED && proxies?.results?.length > 0 && (
+          <StackItem>
+            <Grid>
+              <GridItem span={8}>
+                <Form>
+                  <FormGroup label={__('Select OpenSCAP Capsule')}>
+                    <Select
```
MariaAga: nit, and not needed to change, but using `import { SimpleSelect } from '@patternfly/react-templates';` would have probably simplified the code

webpack/components/HostsIndex/BulkActions/changeOpenscapProxy/BulkChangeOpenscapProxyModal.js
```diff
@@ -0,0 +1,275 @@
...
+  selectAPIResponse,
+} from 'foremanReact/redux/API/APISelectors';
+import {
+  BULK_CHANGE_OPENSCAP_PROXY_KEY,
+  HOSTS_API_PATH,
+  HOSTS_API_REQUEST_KEY,
+  OPENSCAP_PROXIES_KEY,
+} from '../../../OpenscapRemediationWizard/constants';
+
+const buildBulkRequestBody = ({
```
MariaAga: buildBulkRequestBody reimplemented locally — import from foremanReact/components/HostsIndex/BulkActions/helpers

MariaAga (changes requested):
initial review: 
Capsule is for downstream, and gets translated from "Proxy" please only use proxy here.

Going to do some tests now, and then a more in depth review once it gets updated

webpack/components/HostsIndex/BulkActions/changeOpenscapProxy/BulkChangeOpenscapProxyModal.js
Lukshio: I was wondering what will be the best approach, but I decided to not use `FormattedMessage` so we don't rely on external libs.

MariaAga: we use `react-intl` anyway for out translations

app/controllers/concerns/foreman_openscap/api/v2/hosts_bulk_actions_controller_extensions.rb
Lukshio: I tried changing org and location but it does not show count either

sbernhard:
Please have a look at this, too: https://github.com/theforeman/foreman_openscap/pull/615
Might be a valid different approach or a nice addition.

webpack/components/HostsIndex/BulkActions/changeOpenscapProxy/BulkChangeOpenscapProxyModal.js
```diff
@@ -0,0 +1,229 @@
...
+    dispatch(
+      APIActions.get({
+        key: HOSTS_API_REQUEST_KEY,
+        url: foremanUrl(HOSTS_API_PATH),
+      })
+    );
+    handleModalClose();
+  };
+
+  const handleError = () => {
```
sbernhard, line 78: Would it be possible to re-use the error handling from https://github.com/theforeman/foreman/pull/11045 ?

Lukshio: @sbernhard Thanks for the feedback. It is actually solving different issues. In my case, I am using `handleError` for closing the modal and setting state. For `handleSuccess` I am using for refreshing the page and closing modal.

The foreman PR is solving parsing error/sucess messages from the server response and displaying `Toast` by using `addToast`. In my case, I am using directly `success/errorToast` built in `APIActions` so it's actually different codes for different purpose.

webpack/components/HostsIndex/BulkActions/changeOpenscapProxy/BulkChangeOpenscapProxyModal.js
Lukshio: Would you prefer to have it in `FormattedMessage` or I can leave it like this?

webpack/components/HostsIndex/BulkActions/changeOpenscapProxy/BulkChangeOpenscapProxyModal.js
Lukshio: I also added my comments to the PR

webpack/components/HostsIndex/BulkActions/changeOpenscapProxy/BulkChangeOpenscapProxyModal.js
MariaAga: I think you can leave it as is. was mostly sharing info

webpack/components/HostsIndex/BulkActions/changeOpenscapProxy/BulkChangeOpenscapProxyModal.js
```diff
@@ -0,0 +1,229 @@
...
+      variant="small"
+      position="top"
+      actions={modalActions}
+      id="bulk-change-openscap-proxy-modal"
+      key="bulk-change-openscap-proxy-modal"
+      ouiaId="bulk-change-openscap-proxy-modal"
+    >
+      <Stack hasGutter>
+        <StackItem>{descriptionText}</StackItem>
+        {proxyStatus === STATUS.RESOLVED && proxies?.results?.length > 0 && (
```
MariaAga, line 173: Could you add a spinner/skeleton until its resolved?

webpack/components/HostsIndex/BulkActions/changeOpenscapProxy/BulkChangeOpenscapProxyModal.js
MariaAga: I'm starting to think we might need a generic `handleSuccess` as a lot of the bulk actions ignore the perPage param (and probably others) in the data reset - and we can either fix it 1, by 1, or send something similar to `fetchBulkParams` - like `refreshData` in the `ForemanActionsBarContext`

MariaAga (commented):
(midway review as I cant comment on the discussion without clicking "submit review")
Capsule is still mentioned in the code, PR title, and commits.
Also casing should be always `OpenSCAP Proxy` (and not `OpenSCAP proxy`)

webpack/components/HostsIndex/BulkActions/changeOpenscapProxy/BulkChangeOpenscapProxyModal.js
```diff
@@ -0,0 +1,229 @@
...
+      openscap_proxy_id: proxyId,
+    });
+
+    setIsSubmitting(true);
+    dispatch(
+      APIActions.put({
+        key: BULK_CHANGE_OPENSCAP_PROXY_KEY,
+        url: foremanUrl('/api/v2/hosts/bulk/change_openscap_proxy'),
+        handleSuccess,
+        successToast: response => response.data.message,
```
MariaAga, line 97: This doesnt parse the response correctly so the toast just says "success"

app/controllers/concerns/foreman_openscap/api/v2/hosts_bulk_actions_controller_extensions.rb
```diff
@@ -0,0 +1,78 @@
...
+          if failed_hosts.empty?
+            process_response(true, {
+              message: _("OpenSCAP Proxy set to %s") % smart_proxy.name,
+            })
+          else
+            render_error(:bulk_hosts_error, status: :unprocessable_entity,
+                         locals: {
+                           message: n_("Failed to assign OpenSCAP Proxy to %s host",
+                                       "Failed to assign OpenSCAP Proxy to %s hosts",
+                                       failed_hosts.count) % failed_hosts.count,
```
MariaAga: Could the message include also how many were successful/not done? (if I try 60, and see that 55 failed its unclear if the other 5 were skipped or successful)

webpack/components/HostsIndex/BulkActions/changeOpenscapProxy/BulkChangeOpenscapProxyModal.js
Lukshio: I think that at this moment we should finish the changes and schedule some further reactor in cooperation with UX team, unifying the design and code.

MariaAga: Yeah good idea.

webpack/components/HostsIndex/BulkActions/changeOpenscapProxy/BulkChangeOpenscapProxyModal.js
```diff
@@ -0,0 +1,229 @@
...
+    closeModal();
+  };
+
+  const handleSuccess = () => {
+    dispatch(
+      APIActions.get({
+        key: HOSTS_API_REQUEST_KEY,
+        url: foremanUrl(HOSTS_API_PATH),
+      })
+    );
```
MariaAga: I'm hoping to get this in to avoid bugs https://github.com/theforeman/foreman/pull/11110

MariaAga (changes requested):

webpack/components/HostsIndex/BulkActions/changeOpenscapProxy/BulkChangeOpenscapProxyModal.js
Lukshio: It is parsed correctly, but it usually exceeds this rule: https://github.com/theforeman/foreman/blob/7170b302c65d0a318179151f41fb7c699fe01629/webpack/assets/javascripts/react_app/components/ToastsList/helpers.js#L23

thinking of that 60 chars are really low limit to our messages

webpack/components/HostsIndex/BulkActions/changeOpenscapProxy/BulkChangeOpenscapProxyModal.js
Lukshio: Let's wait until it gets merged

webpack/components/HostsIndex/BulkActions/changeOpenscapProxy/BulkChangeOpenscapProxyModal.js
MariaAga: You can wrap it in a `<div>` so it will be sent inside the alert body
image
```js
      addToast({
        type: 'success',
        message: <div>Ex amet voluptate laborum sint ea ut in commodo. Nulla voluptate velit qui est cupidatat ea et commodo ipsum proident adipisicing occaecat. Consequat quis commodo minim nulla ipsum aute ipsum.</div>
      })
```

Lukshio: I tried to avoid using extra `addToast` when `successToast` can be used, but I guess that there is no other option. Then it probably can be aligned with https://github.com/theforeman/foreman/pull/11045 or maybe it should be part of the bigger bulk actions refactor. Maybe we could edit `addToast` to convert title to text once it reach the 60 chars limitation, wdym @MariaAga

MariaAga: >Maybe we could edit addToast to convert title to text once it reach the 60 chars limitation

That would be a good feature

app/controllers/concerns/foreman_openscap/api/v2/hosts_bulk_actions_controller_extensions.rb
```diff
@@ -0,0 +1,91 @@
...
+            failed_hosts << host unless host.save
+          end
+
+          if failed_hosts.empty?
+            process_response(true, {
+              message: _("OpenSCAP Proxy set to %s") % smart_proxy.name,
+            })
+          else
+            total_count = @hosts.count
+            failed_count = failed_hosts.count
```
MariaAga: @hosts.count fires SQL COUNT — Use .size instead on the already-loaded relation.

app/controllers/concerns/foreman_openscap/api/v2/hosts_bulk_actions_controller_extensions.rb
MariaAga, line 1: @ofedoren Could you take a look at the ruby here? 
Should this be under api? then its missing docs, or should it be under controllers

webpack/components/HostsIndex/__tests__/ChangeOpenscapProxyAction.test.js
```diff
@@ -0,0 +1,174 @@
...
+  CHANGE_OPENSCAP_MODAL_ID,
+  OPENSCAP_PROXIES_KEY,
+} from '../../OpenscapRemediationWizard/constants';
+import BulkChangeOpenscapProxyModalScene, {
+  ChangeOpenscapProxyMenuItem,
+} from '../ChangeOpenscapProxyAction';
+
+const { renderWithStore } = rtlHelpers;
+
+jest.mock('foremanReact/common/I18n');
```
MariaAga: why does `jest.mock('foremanReact/common/I18n');` need a mock in the tests?

webpack/components/HostsIndex/BulkActions/changeOpenscapProxy/BulkChangeOpenscapProxyModal.js
```diff
@@ -0,0 +1,243 @@
...
+  Button,
+  Grid,
+  GridItem,
+  Form,
+  FormGroup,
+  Stack,
+  StackItem,
+  Skeleton,
+} from '@patternfly/react-core';
+// eslint-disable-next-line import/no-unresolved
```
MariaAga: Not needed with foreman lint

MariaAga (changes requested):

webpack/components/HostsIndex/BulkActions/changeOpenscapProxy/BulkChangeOpenscapProxyModal.js
Lukshio: Feature introduced here, in favor of that I am bringing back old handler https://github.com/theforeman/foreman/pull/11114

webpack/components/HostsIndex/__tests__/ChangeOpenscapProxyAction.test.js
Lukshio: Removed leftover

Lukshio:
Adapted to https://github.com/theforeman/foreman/pull/11110

app/controllers/concerns/foreman_openscap/api/v2/hosts_bulk_actions_controller_extensions.rb
ofedoren: This looks more like a separate API endpoint instead of actual extension/override. This doesn't change an existing route/action/controller, but adds a brand new endpoint with it's own route, action, etc.

I'd say this should go under regular API endpoints. I think I've seen somewhere similar addition, but from the top of my head, Katello does something like: https://github.com/Katello/katello/blob/master/app/controllers/katello/api/v2/hosts_bulk_actions_controller.rb

Unfortunately I'm a bit swamped with something else to do a proper review, so I'll defer to @adamruzicka to steal a bit of his time 😈

app/controllers/concerns/foreman_openscap/api/v2/hosts_bulk_actions_controller_extensions.rb
```diff
@@ -0,0 +1,104 @@
...
+                message: _("No OpenSCAP Proxy selected."),
+              },
+            }, status: :unprocessable_entity
+          end
+
+          smart_proxy = begin
+            ::SmartProxy.find(openscap_proxy_id)
+          rescue ActiveRecord::RecordNotFound
+            nil
+          end
```
adamruzicka: _nitpick:_
```suggestion
          smart_proxy = ::SmartProxy.find_by(id: openscap_proxy_id)
```

app/controllers/concerns/foreman_openscap/api/v2/hosts_bulk_actions_controller_extensions.rb
```diff
@@ -0,0 +1,104 @@
...
+            }, status: :unprocessable_entity
+          end
+
+          unless smart_proxy.has_feature?('Openscap')
+            return render json: {
+              error: {
+                message: _("The selected proxy does not have the OpenSCAP feature enabled."),
+              },
+            }, status: :unprocessable_entity
+          end
```
adamruzicka: Those could be broken out to ~3 `before_action`s, leaving the actual endpoint method to only deal with the happy path.

adamruzicka (commented):
I'd also vote for a separate controller over extending the existing one as suggested by @ofedoren . Some nitpicks inline

Lukshio:
@adamruzicka I updated the PR based on the comments, ready for re-review

app/controllers/api/v2/compliance/hosts_bulk_actions_controller.rb
```diff
@@ -0,0 +1,101 @@
...
+
+        render json: {
+          :error => {
+            :message => _("No OpenSCAP Proxy selected."),
+          },
+        }, :status => :unprocessable_entity
+      end
+
+      def find_openscap_proxy
+        @smart_proxy = ::SmartProxy.find_by(:id => params[:openscap_proxy_id])
```
adamruzicka: This feels a bit excessive, I'd be fine with following what katello does (either the id is provided and then the resource may (or may not) be found or the id is not provided and then the resource won't be found).

Alternatively, we could rely on strong parameters instead to handle ~90% of this. The error is then slightly less nice, but still understandable.

```suggestion
      def find_openscap_proxy
        @smart_proxy = ::SmartProxy.find_by(:id => params.require(:openscap_proxy_id))
```
\+ the corresponding before_action change + a `rescue_from ActionController::ParameterMissing` near the top to turn this from 500 to 422.

The error then becomes:
```
{
  "error": {"message":"param is missing or the value is empty: openscap_proxy_id"}
}
```

adamruzicka:
```suggestion
        @smart_proxy = ::SmartProxy.authorized.find_by(:id => params[:openscap_proxy_id])
```

app/controllers/api/v2/compliance/hosts_bulk_actions_controller.rb
```diff
@@ -0,0 +1,101 @@
...
+        return if @smart_proxy
+
+        render json: {
+          :error => {
+            :message => _("OpenSCAP Proxy with id %s not found") % params[:openscap_proxy_id],
+          },
+        }, :status => :unprocessable_entity
```
adamruzicka: Any reason for not reusing `render_error`? Applies to muliple places

app/controllers/api/v2/compliance/hosts_bulk_actions_controller.rb
```diff
@@ -0,0 +1,101 @@
...
+          param :ids, Array, :required => false, :desc => N_("List of host ids to exclude and not perform the action on")
+        end
+      end
+
+      api :PUT, "/hosts/bulk/change_openscap_proxy", N_("Assign OpenSCAP Proxy to multiple hosts")
+      param_group :bulk_host_ids
+      param :openscap_proxy_id, :number, :required => true, :desc => N_("ID of the OpenSCAP Proxy to assign to the hosts")
+      def change_openscap_proxy
+        failed_hosts = []
+        @hosts.each do |host|
```
adamruzicka: I know this is consistent with other bulk actions, but a find_each would be a better fit to prevent loading all the hosts at once

```suggestion
        @hosts.find_each do |host|
```

app/controllers/api/v2/compliance/hosts_bulk_actions_controller.rb
```diff
@@ -0,0 +1,101 @@
...
+          host.openscap_proxy = @smart_proxy
+          failed_hosts << host unless host.save
```
adamruzicka: If I'm reading it right, we use `failed_hosts` in two ways. We check counts (L34, L41) and we return the list of their ids. Assuming that's true, could we just collect the ids instead of the full host objects?

adamruzicka (commented):
the permission check is necessary, the rest is optional

app/controllers/api/v2/compliance/hosts_bulk_actions_controller.rb
Lukshio: I just copied it from somewhere else, fixed - unified.

app/controllers/api/v2/compliance/hosts_bulk_actions_controller.rb
Lukshio: Added

Lukshio: Added

app/controllers/api/v2/compliance/hosts_bulk_actions_controller.rb
Lukshio: Updated, thanks for the feedback, I am still learning

app/controllers/api/v2/compliance/hosts_bulk_actions_controller.rb
Lukshio: Changed to ids

app/controllers/api/v2/compliance/hosts_bulk_actions_controller.rb
```diff
@@ -0,0 +1,90 @@
+module Api::V2
+  module Compliance
+    class HostsBulkActionsController < ::Api::V2::BaseController
+      include Api::Version2
```
adamruzicka: `::Api::V2::BaseController` already includes `Api::Version2`
```suggestion
```

adamruzicka (commented):
One last nitpick, oterwise lgtm

app/controllers/api/v2/compliance/hosts_bulk_actions_controller.rb
Lukshio: done, thanks

Lukshio:
Rebase

MariaAga (approved):
Tested it with packit and it looks good to me! 
(@adamruzicka I dont have permissions to merge 😶 )

PR #617, opened 2026-07-21
Fixes #39548 - Fix lint tests, and clean up js code

MariaAga:
While fixing the lint errors, I notice that some come from un-imported/unused files, so deleted them instead.
Lint fix for https://github.com/theforeman/foreman/pull/11008 https://github.com/theforeman/foreman/pull/11098 https://github.com/theforeman/foreman/pull/10968


pondrejk (approved):

webpack/components/LineChart/LineChartHelpers.js
```diff
@@ -1,6 +1,16 @@
...
 import { chart_color_black_500 as chartColorBlack500 } from '@patternfly/react-tokens';
 
+const TIMESTAMP_THRESHOLD = 1e12;
+const MS_PER_SECOND = 1000;
+const DEFAULT_TICK_MID = 0.5;
+const MIN_TICK_STEP = 0.1;
+const EXPONENTIAL_THRESHOLD = 1e21;
+const CLAMP_SCALE_FACTOR = 0.9;
+const HOURS_PER_HALF_DAY = 12;
+const MINUTES_PER_HOUR = 60;
```
adamruzicka: Eh, do we really really have to declare a constant saying that there are 60 minutes in an hour _in every single plugin we have_? I mean I get where it's coming from, but it feels a bit silly

MariaAga: I mostly dont want someone to delete a small constant from core because they think no one uses them, and cause much damage. It seems like a simple way to ensure that. As 1 hour is a fact generally, its not like core decides what is an hour. (unlike status names for example)

Lukshio: @adamruzicka I had same question, and now I think that it will be safer do redefine it in plugins.

adamruzicka: I can't say I'm convinced.

While the argument of "what if we delete the constant from core" is appealing from the developer's point of view, it also sounds a lot like "but then we won't be able to freely change what is essentially a public interface".

> As 1 hour is a fact generally, its not like core decides what is an hour. (unlike status names for example)

This goes both ways. Neither foreman nor plugins decide what http 401 means, yet this one gets imported, while the others do not.

sbernhard: I totally agree with @adamruzicka. 
Just had a look, there are some libraries like https://github.com/date-fns/date-fns/blob/4098115cf705e3af7f663d8e5b0686e39a9f478a/pkgs/core/src/constants/index.ts#L133 - but it would be another npm package we need to maintain...

MariaAga: Thanks @adamruzicka & @sbernhard , those are good points, I will update my PRs to import from core. 
And as you said, we want to avoid more npm packages to maintain, so I'll just keep the data in core.

adamruzicka (changes requested):
_To formalize https://github.com/theforeman/foreman_openscap/pull/617/changes#r3645773200_ : Constants available in foreman core should not be duplicated.

MariaAga:
depends on https://github.com/theforeman/foreman/pull/11120

adamruzicka (approved):
I'm rather fond of the stat and the tests don't seem to mind.

adamruzicka:
@MariaAga could you please squash the last two commits together?

MariaAga:
squashed, tests are :green_apple:

adamruzicka (approved):

adamruzicka:
Thank you @MariaAga !

PR #618, opened 2026-07-29
Fixes #39566 - add test_setup file that imports foremanJSTestSetup

andreilakatos:
Fixes #39566 - add test_setup file that imports foremanJSTestSetup

### DIFF
--- a/webpack/test_setup.js
+++ b/webpack/test_setup.js
@@ -1,0 +1,1 @@
+import 'foremanJSTestSetup';


ofedoren:
Not against that, but wonder: why do we need that exactly? It's not mentioned neither in RM nor Jira ticket...

andreilakatos:
> Not against that, but wonder: why do we need that exactly? It's not mentioned neither in RM nor Jira ticket...

This is the ticket explaining it:

https://redhat.atlassian.net/browse/SAT-48340
`Foreman core has a test setup and plugins should reuse that instead of duplicating it. This aims to make it easier to maintain.`

ofedoren:
Thanks, @andreilakatos !

Ruby test failures are unrelated, but JS ones are :/

ofedoren:
Apologies, I'm dumb (or just blind): all the JS failures should be fixed by https://github.com/theforeman/foreman_openscap/pull/617 (linting, constants, etc), so we'd rather get that one in first, then rebase this one and merge if nothing's weird.

andreilakatos:
> Apologies, I'm dumb (or just blind): all the JS failures should be fixed by #617 (linting, constants, etc), so we'd rather get that one in first, then rebase this one and merge if nothing's weird.

Happens to everyone, lol

adamruzicka (approved):

adamruzicka:
Thank you @andreilakatos !

PR #619, opened 2026-08-13
harden: found an unscoped `find( in compliance_hosts_controller.rb...

anupamme:
## Summary
Harden input handling in `app/controllers/compliance_hosts_controller.rb` (flagged by semgrep).

## Vulnerability
| Field | Value |
|-------|-------|
| **ID** | ruby.rails.security.brakeman.check-unscoped-find.check-unscoped-find |
| **Severity** | HIGH |
| **Scanner** | semgrep |
| **Rule** | `ruby.rails.security.brakeman.check-unscoped-find.check-unscoped-find` |
| **File** | `app/controllers/compliance_hosts_controller.rb:3` |
| **Assessment** | Defensive hardening |

**Description**: Found an unscoped `find(...)` with user-controllable input. If the ActiveRecord model being searched against is sensitive, this may lead to Insecure Direct Object Reference (IDOR) behavior and allow users to read arbitrary records. Scope the find to the current user, e.g. `current_user.accounts.find(params[:id])`.

## Threat Model Context

This controller appears to be publicly accessible. This is a Node.js library - vulnerabilities affect downstream consumers who use this package.

## Changes
- `app/controllers/compliance_hosts_controller.rb`

## Behavior Preservation
The change is scoped to 1 file on the vulnerable path; it only tightens handling of untrusted input and leaves valid inputs unaffected.

---
*This patch removes an exploit primitive — a code pattern that, while not independently exploitable today, could be chained with other weaknesses by automated exploit-development tooling. Proactive removal of such primitives raises the bar against increasingly capable automated attack tools.*

---
*Automated security fix by [OrbisAI Security](https://orbisappsec.com)*


adamruzicka (commented):
Thank you for the patch.

Instead of changing what we do in show, could we instead include the `FindCommon` concern and delegate the lookup to a `find_resource` before_action?

anupamme:
Review comments addressed. Pls review.

adamruzicka:
Looks reasonable, let's see what CI thinks.

On a more general note, this project uses redmine (https://projects.theforeman.org/projects/foreman_openscap) for issue tracking and commit messages have to be shaped like `Fixes #<issue-id> - <description>`.

PR #623, opened 2026-09-14
Fixes #39761 - Add un/assign compliance policy bulk action to hosts

kmalyjur:
**QUESTION**: As you can see, the new UI is now missing the table with hosts - do you think it's important and should be added? Also, in the new UI, after submitting the modal, the hosts are still selected in the table, so the "Remember hosts selection..." tick is not needed.

Change in the behaviour when assigning a policy that is already assigned:
--- Legacy: `Error: cannot assign to centos-vm.usersys.redhat.com, all assigned policies must be deployed in the same way, check 'deploy by' for each assigned policy`
--- Now: `Successfully assigned compliance policy to selected hosts`

Legacy UI:
image

New UI:
image


webpack/components/BulkActions/compliancePolicy/BulkCompliancePolicyMenuItems.js
```diff
@@ -0,0 +1,39 @@
...
+const BulkCompliancePolicyMenuItems = () => {
+  const { selectedCount } = useContext(ForemanHostsIndexActionsBarContext);
+
+  const handleAssignClick = () => {
+    window.dispatchEvent(new CustomEvent('openscap:openAssignModal'));
+  };
+
+  const handleUnassignClick = () => {
+    window.dispatchEvent(new CustomEvent('openscap:openUnassignModal'));
+  };
```
Lukshio: Why did you chose to use this instead of `openBulkModal` as other modals use?

webpack/components/BulkActions/compliancePolicy/BulkCompliancePolicyModal.js
```diff
@@ -0,0 +1,285 @@
...
+        <Text ouiaId="bulk-compliance-policy-warning">{warningMessage}</Text>
+      </TextContent>
+
+      {policiesStatus === STATUS.PENDING && (
+        <div style={{ textAlign: 'center', padding: '20px' }}>
+          <Spinner size="lg" aria-label={__('Loading policies')} />
+          <Text ouiaId="loading-policies-text" style={{ marginTop: '10px' }}>
+            {__('Loading policies...')}
+          </Text>
+        </div>
```
Lukshio: Use `<Skeleton>` instead or `<Flex>` instead of custom css div

webpack/components/BulkActions/compliancePolicy/index.js
Lukshio, line 1: Do not use `index.js` files https://github.com/theforeman/foreman/blob/07723412193b5c5fb5c8e4f4a807bb39112f528f/developer_docs/pf3-to-pf5-migration-guide.asciidoc?plain=1#L27

webpack/components/BulkActions/compliancePolicy/BulkCompliancePolicyWrapper.js
```diff
@@ -0,0 +1,40 @@
...
+    window.addEventListener('openscap:openUnassignModal', handleOpenUnassign);
+
+    return () => {
+      window.removeEventListener('openscap:openAssignModal', handleOpenAssign);
+      window.removeEventListener(
+        'openscap:openUnassignModal',
+        handleOpenUnassign
+      );
+    };
+  }, []);
```
Lukshio: This is redo of `useBulkModalOpen` from `bulkModalStateHelper`

webpack/index.js
```diff
@@ -3,6 +3,7 @@ import componentRegistry from 'foremanReact/components/componentRegistry';
 import RuleSeverity from './components/RuleSeverity';
 import OpenscapRemediationWizard from './components/OpenscapRemediationWizard';
 import LineChart from './components/LineChart';
+import './global_index';
```
Lukshio: global_index is loaded every time, this import can be dropped

webpack/components/BulkActions/compliancePolicy/actions.js
```diff
@@ -0,0 +1,54 @@
...
+    `/compliance/policies/update_multiple_hosts?${queryParams.toString()}`
+  );
+
+  return APIActions.post({
+    url,
+    key: BULK_ASSIGN_POLICY_KEY,
+    params: { policy: { id: policy_id } },
+    successToast: () =>
+      __('Successfully assigned compliance policy to selected hosts'),
+    errorToast: ({ message }) => message,
```
Lukshio: `onError` use `addToast`, so the error toast can be duplicated because `errorToast` is also used

webpack/components/BulkActions/compliancePolicy/actions.js
```diff
@@ -0,0 +1,54 @@
...
+
+  return APIActions.post({
+    url,
+    key: BULK_UNASSIGN_POLICY_KEY,
+    params: { policy: { id: policy_id } },
+    successToast: () =>
+      __('Successfully unassigned compliance policy from selected hosts'),
+    errorToast: ({ message }) => message,
+    handleSuccess: onSuccess,
+    handleError: onError,
```
Lukshio: same as above

webpack/components/BulkActions/compliancePolicy/index.js
```diff
@@ -0,0 +1,40 @@
...
+    locationId,
+    refreshTableData,
+  } = useContext(ForemanActionsBarContext);
+
+  return (
+    <BulkCompliancePolicyModal
+      key={`bulk-compliance-policy-modal-${mode}`}
+      selectAllHostsMode={selectAllHostsMode}
+      selectedCount={selectedCount}
+      selectedResults={selectedResults}
```
Lukshio: I didn't found any usage

webpack/components/BulkActions/compliancePolicy/BulkCompliancePolicyMenuItems.js
Lukshio, line 1: I would expect some permissons check for `edit_hosts` before showing it in menu

webpack/components/BulkActions/compliancePolicy/BulkCompliancePolicyModal.js
```diff
@@ -0,0 +1,285 @@
...
+        'The selected policy will be unassigned from the selected hosts. Some hosts may not be assigned to this policy.'
+      );
+
+  return (
+    <Modal
+      isOpen={isOpen}
+      onClose={handleModalClose}
+      onEscapePress={handleModalClose}
+      title={title}
+      width="50%"
```
Lukshio: Use `variant` instead, this will be different on each screen

Lukshio (changes requested):

kmalyjur:
@Lukshio thank you, updated. I'm just unsure about the permissions. (No component in the codebase checks the `edit_hosts` in the frontend.)
Also, what do you think about "As you can see, the new UI is now missing the table with hosts - do you think it's important and should be added?" We can ask Maria from UX about that.

kmalyjur:
`fix false success` commit:
Controller returned HTTP redirects for both success and errors and showed success even when validation failed.
Solution: Detect AJAX requests (request.xhr?) and return JSON with the correct status code instead of redirects.

Lukshio (changes requested):
> @Lukshio thank you, updated. I'm just unsure about the permissions. (No component in the codebase checks the `edit_hosts` in the frontend.) 

We use permission check often e.g. 
https://github.com/theforeman/foreman/blob/c74b70ed8ea3d2fbdb206a636d9c528578eddc01/webpack/assets/javascripts/react_app/components/HostsIndex/index.js#L52

> Also, what do you think about "As you can see, the new UI is now missing the table with hosts - do you think it's important and should be added?" We can ask Maria from UX about that.

All updated modals are without tables, so I would leave it like that 

>fix false success commit:
Controller returned HTTP redirects for both success and errors and showed success even when validation failed.
Solution: Detect AJAX requests (request.xhr?) and return JSON with the correct status code instead of redirects.

I would introduce new API controller, because later the legacy page will be deleted and it will be cleaner variant.

webpack/components/BulkActions/compliancePolicy/BulkCompliancePolicyMenuItems.js
```diff
@@ -1,34 +1,37 @@
...
 
 const BulkCompliancePolicyMenuItems = () => {
   const { selectedCount } = useContext(ForemanHostsIndexActionsBarContext);
+  const userPermissions = useForemanPermissions();
 
-  const handleAssignClick = () => {
-    window.dispatchEvent(new CustomEvent('openscap:openAssignModal'));
-  };
+  const hasPermission =
+    userPermissions.has('assign_policies') && userPermissions.has('edit_hosts');
```
Lukshio: Please use consts from core

Lukshio (changes requested):

webpack/components/BulkActions/compliancePolicy/BulkCompliancePolicyMenuItems.js
kmalyjur: oops I meant to change it

webpack/components/BulkActions/compliancePolicy/actions.js
```diff
@@ -0,0 +1,31 @@
...
+import {
+  POLICIES_KEY,
+  BULK_ASSIGN_POLICY_KEY,
+  BULK_UNASSIGN_POLICY_KEY,
+} from './constants';
+
+export const fetchPolicies = () =>
+  APIActions.get({
+    url: foremanUrl('/api/v2/compliance/policies'),
+    key: POLICIES_KEY,
```
Lukshio, line 12:
```suggestion
    key: POLICIES_KEY,
    params: { per_page: 'all' },
```

i would include `all`, I think that otherwise it will use `per_page` from user settings

app/controllers/api/v2/compliance/hosts_bulk_actions_controller.rb
Lukshio, line 1: Add please new test cases

webpack/components/BulkActions/compliancePolicy/BulkCompliancePolicyModal.js
```diff
@@ -0,0 +1,299 @@
...
+    ? __('Assign policy')
+    : __('Unassign policy');
+
+  const modalActions = [
+    <Button
+      key="confirm"
+      ouiaId="bulk-compliance-policy-modal-confirm-button"
+      variant="primary"
+      onClick={handleConfirm}
+      isDisabled={policyId === ''}
```
Lukshio, line 148: Add please condition to prevent double submit

webpack/components/BulkActions/compliancePolicy/BulkCompliancePolicyModal.js
```diff
@@ -0,0 +1,299 @@
...
+    policies?.results?.find(p => p.id.toString() === id)?.name ||
+    __('Select a policy');
+
+  const toggle = toggleRef => (
+    <MenuToggle
+      ref={toggleRef}
+      ouiaId="bulk-compliance-policy-toggle"
+      onClick={onToggleClick}
+      isExpanded={policySelectOpen}
+      style={{ width: '500px' }}
```
Lukshio, line 89: Do not use hardcoded values

Lukshio (changes requested):

